PR #6, opened 2026-04-19
Added shared Ollama process ownership with health monitor

srvsngh99:
## Summary

- Mini Krill now owns the Ollama process lifecycle for the colony (DeepKrill, Mini Krill, Lab Krill)
- Added `EnsureRunning()` method and `weStartedIt` tracking so we only restart processes we started
- Background health monitor (15s probe interval, exponential backoff restart, max 5 crashes)
- New `minikrill ollama ensure` command: install + start + pull in one shot for scripting
- `dive` command now auto-starts Ollama when `auto_start: true` (was a dead flag, now wired)
- `surface`/exit keeps Ollama running so other bots aren't affected
- Default model updated to `gemma4:e2b` across config and code defaults

## Test plan

- [ ] `minikrill ollama ensure` installs, starts, pulls gemma4:e2b, prints "ready"
- [ ] `minikrill dive` auto-starts Ollama before LLM init
- [ ] Kill Ollama process manually, verify health monitor restarts within 15s
- [ ] `minikrill surface` stops Mini Krill but Ollama keeps running
- [ ] `minikrill ollama stop` actually kills the Ollama process
- [ ] `minikrill ollama status` correctly reports running/stopped state

### DIFF
--- a/cmd/minikrill/commands.go
+++ b/cmd/minikrill/commands.go
@@ -257,6 +257,18 @@ var diveCmd = &cobra.Command{
 
 // diveDaemon re-execs the dive command as a detached background process.
 func diveDaemon() error {
+	// Check if a daemon is already running
+	pidFile := filepath.Join(config.DataDir(), "krill.pid")
+	if data, err := os.ReadFile(pidFile); err == nil {
+		if pid, err := strconv.Atoi(strings.TrimSpace(string(data))); err == nil {
+			if proc, err := os.FindProcess(pid); err == nil {
+				if proc.Signal(syscall.Signal(0)) == nil {
+					return fmt.Errorf("Mini Krill is already diving (PID %d). Run 'minikrill surface' first", pid)
+				}
+			}
+		}
+	}
+
 	exe, err := os.Executable()
 	if err != nil {
 		return fmt.Errorf("resolve executable: %w", err)
@@ -290,7 +302,6 @@ func diveDaemon() error {
 	// Detach - the parent does not wait for the child.
 	_ = child.Process.Release()
 
-	pidFile := filepath.Join(config.DataDir(), "krill.pid")
 	_ = os.WriteFile(pidFile, []byte(strconv.Itoa(child.Process.Pid)), 0644)
 
 	printBanner()
--- a/cmd/minikrill/commands.go
+++ b/cmd/minikrill/commands.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"os/exec"
 	"os/signal"
 	"path/filepath"
 	"runtime"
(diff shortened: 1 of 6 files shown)


internal/ollama/manager.go
```diff
@@ -370,3 +390,106 @@ func (m *OllamaManager) isHealthy(ctx context.Context) bool {
...
 	return resp.StatusCode == http.StatusOK
 }
+
+// HasModel checks if a model is available locally.
+func (m *OllamaManager) HasModel(ctx context.Context, model string) bool {
+	models, err := m.ListModels(ctx)
+	if err != nil {
+		return false
+	}
+	base := strings.Split(strings.ToLower(model), ":")[0]
```
srvsngh99: **Bug: Tag matching is too loose — will skip pulling the needed model variant.**

This strips the tag and only compares base names. If `gemma4:latest` is already installed and you check for `gemma4:e2b`, `HasModel` returns `true` and `ensure` silently skips the pull.

Should do an exact match, normalizing the implicit `:latest` tag:
```go
func (m *OllamaManager) HasModel(ctx context.Context, model string) bool {
	models, err := m.ListModels(ctx)
	if err != nil {
		return false
	}
	normalize := func(s string) string {
		s = strings.ToLower(s)
		if !strings.Contains(s, ":") {
			s += ":latest"
		}
		return s
	}
	target := normalize(model)
	for _, mi := range models {
		if normalize(mi.Name) == target {
			return true
		}
	}
	return false
}
```

srvsngh99 (commented):
Review: Good feature overall — lifecycle management, health monitor with backoff, `ensure` command, and daemon mode are well-structured. A few issues to address before merging.

Also: `cmd/minikrill/cmd_ollama.go:88` still suggests `llama3.2` as the example model to pull, but the default has been changed to `gemma4:e2b`. Should update to match.

internal/ollama/manager.go
```diff
@@ -134,12 +137,28 @@ func (m *OllamaManager) installViaScript(ctx context.Context) error {
 // Lifecycle
 // ---------------------------------------------------------------------------
 
+// EnsureRunning checks if Ollama is healthy and starts it if not.
+// Intended as the single entry point for "make sure Ollama is up".
+func (m *OllamaManager) EnsureRunning(ctx context.Context) error {
+	if m.isHealthy(ctx) {
+		log.Debug("ollama is already running")
+		m.mu.Lock()
+		m.weStartedIt = false
```
srvsngh99: **Bug: Resets ownership even if we previously started Ollama.**

If Mini Krill started Ollama (`weStartedIt = true`) and something later calls `EnsureRunning` while Ollama is still healthy, this unconditionally sets `weStartedIt = false`. The health monitor would then refuse to restart on a future crash ("not managed by us — will not restart").

Same issue in `Start()` at line 160.

The early-return path should preserve existing ownership:
```go
if m.isHealthy(ctx) {
	log.Debug("ollama is already running")
	// Don't reset weStartedIt — we may already own this process
	return nil
}
```

cmd/minikrill/commands.go
```diff
@@ -225,6 +255,53 @@ var diveCmd = &cobra.Command{
 	},
 }
 
+// diveDaemon re-execs the dive command as a detached background process.
+func diveDaemon() error {
```
srvsngh99, line 259: **Bug: No duplicate-daemon guard.**

Running `minikrill dive` twice starts a second daemon and overwrites `krill.pid`. The first daemon becomes orphaned — `surface` can only kill the second one.

Should check if `krill.pid` already points to a live process:
```go
pidFile := filepath.Join(config.DataDir(), "krill.pid")
if data, err := os.ReadFile(pidFile); err == nil {
	if pid, err := strconv.Atoi(strings.TrimSpace(string(data))); err == nil {
		if proc, err := os.FindProcess(pid); err == nil {
			if proc.Signal(syscall.Signal(0)) == nil {
				return fmt.Errorf("Mini Krill is already diving (PID %d). Run 'minikrill surface' first", pid)
			}
		}
	}
}
```

internal/ollama/manager.go
```diff
@@ -370,3 +390,106 @@ func (m *OllamaManager) isHealthy(ctx context.Context) bool {
...
+	consecutiveCrashes := 0
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-time.After(probeInterval):
+		}
+
+		if m.isHealthy(ctx) {
```
srvsngh99, line 459: **Nit: Shutdown race — unnecessary restart attempt on context cancellation.**

The `select` can wake from `time.After` at the same instant the context is cancelled. Then `isHealthy(ctx)` gets a cancelled context, returns false, and triggers a pointless restart attempt (which also fails). Not harmful but noisy in logs.

Adding a guard after the select:
```go
case <-time.After(probeInterval):
}
if ctx.Err() != nil {
	return
}
```

PR #10, opened 2026-04-30
Replace numbered provider list with interactive arrow-key selector

srvsngh99:
## Summary
- Adds a reusable Bubble Tea-based `promptSelect` component (`cmd/minikrill/prompt.go`) for arrow-key navigable selection lists
- Replaces the `[1] [2] [3]` numbered text input in `minikrill init` with the interactive selector — navigate with ↑/↓ or j/k, press Enter to select
- No new dependencies — uses the existing `charmbracelet/bubbletea` dependency

## Test plan
- [ ] Run `minikrill init` and verify the provider list is navigable with arrow keys
- [ ] Confirm Enter selects the highlighted provider and continues the setup flow
- [ ] Confirm Ctrl+C cancels gracefully
- [ ] Verify subsequent prompts (Y/N confirms, model input, bot tokens) still work after the selector exits

### DIFF
--- a/cmd/minikrill/commands.go
+++ b/cmd/minikrill/commands.go
@@ -132,16 +132,13 @@ var initCmd = &cobra.Command{
 
 			if !localSet[selectedModel] {
 				fmt.Printf(cDim+"  Pulling %s in background..."+cReset+"\n", selectedModel)
+				fmt.Println(cDim + "  Run " + cReset + "minikrill doctor" + cDim + " to check pull status." + cReset)
+				ctx := cmd.Context()
 				go func(model string) {
-					if err := mgr.EnsureRunning(context.Background()); err != nil {
-						fmt.Printf("\n"+cRed+"  Background pull: could not start Ollama: %v"+cReset+"\n", err)
+					if err := mgr.EnsureRunning(ctx); err != nil {
 						return
 					}
-					if err := mgr.Pull(context.Background(), model); err != nil {
-						fmt.Printf("\n"+cRed+"  Background pull of %s failed: %v"+cReset+"\n", model, err)
-					} else {
-						fmt.Printf("\n"+cGreen+"  Model %s is ready!"+cReset+"\n", model)
-					}
+					_ = mgr.Pull(ctx, model)
 				}(selectedModel)
 			}
 		case 1: // Codex
--- a/cmd/minikrill/commands.go
+++ b/cmd/minikrill/commands.go
@@ -49,14 +49,20 @@ var initCmd = &cobra.Command{
 
 		fmt.Println(cBCyan + "  Choose your LLM provider:" + cReset)
 		fmt.Println()
-		fmt.Println("    " + cBGreen + "[1]" + cReset + " Ollama        " + cDim + "local, free, private; pulls gemma3:4b" + cReset)
-		fmt.Println("    " + cBBlue + "[2]" + cReset + " Codex         " + cDim + "optional; ChatGPT subscription via official Codex CLI" + cReset)
-		fmt.Println("    " + cBBlue + "[3]" + cReset + " Claude Code   " + cDim + "optional; Claude Pro/Max via official Claude CLI" + cReset)
-		fmt.Println()
-		choice := ask(scanner, cCyan+"  > "+cReset)
+		providerIdx, err := promptSelect([]selectItem{
+			{label: "Ollama", desc: "local, free, private; pulls gemma3:4b"},
+			{label: "Codex", desc: "ChatGPT subscription via official Codex CLI"},
+			{label: "Claude Code", desc: "Claude Pro/Max via official Claude CLI"},
+		})
+		if err != nil {
+			return err
+		}
+		if providerIdx < 0 {
+			return nil
+		}
 
-		switch choice {
-		case "2", "codex", "chatgpt":
+		switch providerIdx {
+		case 1: // Codex
 			cfg.LLM.Provider = "codex"
 			cfg.LLM.Model = "auto"
 			fmt.Println()
@@ -74,7 +80,7 @@ var initCmd = &cobra.Command{
 					_ = cmd.Run()
 				}
 			}
-		case "3", "claude", "claude code", "claude-code":
+		case 2: // Claude Code
 			cfg.LLM.Provider = "claude"
 			cfg.LLM.Model = "auto"
 			fmt.Println()
@@ -92,7 +98,7 @@ var initCmd = &cobra.Command{
 					_ = cmd.Run()
 				}
 			}
-		default:
+		default: // Ollama (index 0)
 			cfg.LLM.Provider = "ollama"
 			fmt.Println()
 			mgr := ollama.NewManager(cfg.Ollama)
--- a/cmd/minikrill/commands.go
+++ b/cmd/minikrill/commands.go
@@ -58,10 +58,92 @@ var initCmd = &cobra.Command{
 			return err
 		}
 		if providerIdx < 0 {
+			fmt.Println(cDim + "  Setup cancelled." + cReset)
 			return nil
 		}
 
 		switch providerIdx {
+		case 0: // Ollama
+			cfg.LLM.Provider = "ollama"
+			fmt.Println()
+			mgr := ollama.NewManager(cfg.Ollama)
+			if !mgr.IsInstalled() {
+				ans := ask(scanner, cYellow+"  Ollama not found."+cReset+" Install now? "+cDim+"[Y/n]"+cReset+" ")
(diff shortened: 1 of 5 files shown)


srvsngh99 (commented):
## Review Comments

Solid UX improvement overall. A few items to fix before merging — see inline comments.

cmd/minikrill/prompt.go
```diff
@@ -0,0 +1,84 @@
...
+			}
+		case "down", "j":
+			if m.cursor < len(m.items)-1 {
+				m.cursor++
+			}
+		case "enter":
+			m.chosen = m.cursor
+			m.quitting = true
+			return m, tea.Quit
+		case "ctrl+c":
```
srvsngh99: `esc` should also cancel — users unfamiliar with terminal apps won't know to reach for Ctrl+C. Add `"esc"` to this case:
```go
case "ctrl+c", "esc":
```

cmd/minikrill/prompt.go
```diff
@@ -0,0 +1,84 @@
...
+		if m.chosen >= 0 && m.chosen < len(m.items) {
+			return fmt.Sprintf("    %s▸ %s%s\n", cBCyan, m.items[m.chosen].label, cReset)
+		}
+		return ""
+	}
+
+	var b strings.Builder
+	for i, item := range m.items {
+		if i == m.cursor {
+			b.WriteString(fmt.Sprintf("    %s▸ %-14s %s%s\n",
```
srvsngh99: Hardcoded `%-14s` is tuned for the current longest label ("Claude Code" = 11 chars). If a label ever exceeds 14 chars the columns will misalign. Compute max width dynamically:
```go
maxLen := 0
for _, item := range m.items {
    if len(item.label) > maxLen { maxLen = len(item.label) }
}
fmtStr := fmt.Sprintf("    %%s▸ %%-%ds %%s%%s\n", maxLen+2)
```

srvsngh99:
### `commands.go` — two items to fix

**1. Fragile `default` catching Ollama (line 101)**
Using `default` to handle index 0 (Ollama) is fragile — if the slice is reordered or a provider is added, `default` silently absorbs the wrong index. Use an explicit `case 0:` and make `default` return an error:
```go
case 0: // Ollama
    ...
default:
    return fmt.Errorf("unexpected provider index: %d", providerIdx)
```

**2. Silent exit on cancel (line 60–62)**
When Ctrl+C is pressed, the user is dumped back to the shell with no feedback. Add a message:
```go
if providerIdx < 0 {
    fmt.Println(cDim + "  Setup cancelled." + cReset)
    return nil
}
```

cmd/minikrill/prompt.go
```diff
@@ -0,0 +1,92 @@
...
+	for i, item := range m.items {
+		if i == m.cursor {
+			b.WriteString(fmt.Sprintf("    %s▸ %-*s %s%s\n",
+				cBCyan, pad, item.label, item.desc, cReset))
+		} else {
+			b.WriteString(fmt.Sprintf("    %s  %-*s %s%s\n",
+				cDim, pad, item.label, item.desc, cReset))
+		}
+	}
+	b.WriteString("\n    " + cDim + "↑/↓ navigate • enter select" + cReset)
```
srvsngh99: **Nit:** The help text mentions `↑/↓` and `enter` but not `esc`, even though `esc` was explicitly added for discoverability. Consider:
```go
b.WriteString("\n    " + cDim + "↑/↓ navigate • enter select • esc cancel" + cReset)
```

cmd/minikrill/commands.go
```diff
@@ -49,14 +49,102 @@ var initCmd = &cobra.Command{
...
+				if selectedModel == "" {
+					selectedModel = cfg.Ollama.DefaultModel
+				}
+			}
+			cfg.LLM.Model = selectedModel
+			cfg.Ollama.DefaultModel = selectedModel
+
+			if !localSet[selectedModel] {
+				fmt.Printf(cDim+"  Pulling %s in background..."+cReset+"\n", selectedModel)
+				go func(model string) {
```
srvsngh99, line 137: **Medium:** This background goroutine writes to stdout (`fmt.Printf`) while the main goroutine continues through the Telegram/Discord prompts. Output can interleave:
```
  Enable Telegram bot? [y/N]
  Model gemma3:4b is ready!    ← arrives mid-prompt
```
Consider either:
- Suppressing the print and adding a note like "run `minikrill doctor` to check pull status"
- Using a `sync.WaitGroup` to print the result after config is saved
- Or at minimum, deferring the message to after the wizard completes

srvsngh99 (commented):
## Review

Good UX improvement — the interactive selector is well-built and the model list logic (local-first, then suggestions, then custom) is smart. Prior review feedback has all been addressed.

**Blocking:** CI lint is red — `isConfirmation` in `main.go:227` is now unused (removed its only caller), and pre-existing `errcheck` failures in `conversations_test.go` need `_ =` or proper assertions. Both must be fixed before merge.

cmd/minikrill/commands.go
```diff
@@ -49,14 +49,102 @@ var initCmd = &cobra.Command{
...
+					if err := mgr.EnsureRunning(context.Background()); err != nil {
```
srvsngh99: **Low:** `context.Background()` here means the pull won't cancel if the user hits Ctrl+C during init. Since the process exits anyway this is mostly cosmetic, but if you ever add cleanup logic it could be a surprise. A derived context from the command would be cleaner.

cmd/minikrill/commands.go
```diff
@@ -49,14 +49,99 @@ var initCmd = &cobra.Command{
...
+		if err != nil {
+			return err
+		}
+		if providerIdx < 0 {
+			fmt.Println(cDim + "  Setup cancelled." + cReset)
+			return nil
+		}
+
+		switch providerIdx {
+		case 0: // Ollama
```
srvsngh99, line 66: **Blocking (lint):** The old `default` branch used `isConfirmation()` from `main.go:227`. That function is now unused since this PR removed its only call site. Delete `isConfirmation` from `main.go` to fix the `unused` lint error.

srvsngh99 (commented):
## Re-review

All 3 previous inline comments have been addressed in the latest commit — nice work:
- ✅ Help text now shows `esc cancel`
- ✅ Background pull no longer prints to stdout; directs user to `minikrill doctor`
- ✅ Uses `cmd.Context()` instead of `context.Background()`

**Still blocking: CI lint is red** with 2 remaining issues:

1. **`isConfirmation` is unused** (`cmd/minikrill/main.go:227`) — this function's only caller was removed by this PR. Delete the function.
2. **`errcheck` in `conversations_test.go:72-74`** — pre-existing, but still blocking the merge. Three unchecked `store.SaveTurn()` calls need either `_ =` or proper error assertions.

Everything else looks good. Once lint is green this is ready to merge.

srvsngh99:
### Remaining lint fixes to unblock merge

Two changes needed to get CI green:

**1. Delete unused `isConfirmation`** — `cmd/minikrill/main.go:227`
```go
// Remove this entire function — its only caller was removed by this PR
func isConfirmation(s string) bool {
	s = strings.ToLower(s)
	return s == "yes" || s == "y" || s == "ok" || s == "sure" || s == "yep"
}
```

**2. Fix `errcheck` in `internal/brain/conversations_test.go:72-74`**
```diff
- store.SaveTurn("cli", "user", "cli message")
- store.SaveTurn("telegram", "user", "telegram message")
- store.SaveTurn("cli", "assistant", "cli response")
+ _ = store.SaveTurn("cli", "user", "cli message")
+ _ = store.SaveTurn("telegram", "user", "telegram message")
+ _ = store.SaveTurn("cli", "assistant", "cli response")
```

Everything else LGTM — ready to merge once lint is green.

srvsngh99 (commented):
## Re-review (round 3)

Progress since last review:
- ✅ `isConfirmation` deleted — `unused` lint error resolved

**PR diff itself looks clean.** All prior feedback has been addressed. The only remaining CI blocker is the pre-existing `errcheck` in `conversations_test.go:72-74` — three unchecked `store.SaveTurn()` calls. This is not introduced by this PR, but it blocks the lint gate.

Quick fix:
```diff
- store.SaveTurn("cli", "user", "cli message")
- store.SaveTurn("telegram", "user", "telegram message")
- store.SaveTurn("cli", "assistant", "cli response")
+ _ = store.SaveTurn("cli", "user", "cli message")
+ _ = store.SaveTurn("telegram", "user", "telegram message")
+ _ = store.SaveTurn("cli", "assistant", "cli response")
```

Once that's in, all checks should pass and this is good to merge.

PR #12, opened 2026-04-30
Fix chat input bar visibility and tab navigation

srvsngh99:
## Summary
- **Chat input bar invisible**: `resizeViews()` hardcoded 10 lines of overhead for header/tabs/footer, but the ASCII art header takes 14 lines (full mode). This over-allocated body height by ~11 lines, pushing the input bar off-screen. Fixed by computing body height identically to `View()` — rendering actual header/tabbar/footer and counting lines.
- **Tab switching stuck on Chat tab**: Tab/Shift+Tab were blocked when chat input was focused, trapping users. Now Tab/Shift+Tab always switch tabs regardless of input focus. Arrow keys and number keys still require Esc to blur first (needed for cursor movement and typing).
- **Escape key handling**: Added explicit Esc handler to blur chat input and documented it in the Help tab.

## Test plan
- [ ] Run `minikrill chat` — verify the chat input bar with cyan border is visible at the bottom
- [ ] Type a message and press Enter — verify text is visible while typing and message sends
- [ ] Press Tab/Shift+Tab from any tab — verify it cycles through Dashboard, Chat, Logs, Help
- [ ] On Chat tab with input focused, press Tab — verify it switches to the next tab
- [ ] Press Esc on Chat tab — verify input blurs (cursor disappears)
- [ ] Resize terminal window — verify input bar stays visible at all sizes
- [ ] Check Help tab — verify Esc shortcut is listed

### DIFF
--- a/internal/tui/app.go
+++ b/internal/tui/app.go
@@ -34,6 +34,9 @@ const (
 	TabChat      = 1
 	TabLogs      = 2
 	TabHelp      = 3
+
+	// layoutMargins is the vertical breathing room reserved between layout sections.
+	layoutMargins = 3
 )
 
 // ---------------------------------------------------------------------------
@@ -152,16 +155,7 @@ func (a *App) View() string {
 	body := a.renderBody()
 	footer := a.renderFooter()
 
-	// Calculate body height: total - header - tabs - footer - margins
-	headerLines := strings.Count(header, "\n") + 1
-	tabLines := strings.Count(tabBar, "\n") + 1
-	footerLines := strings.Count(footer, "\n") + 1
-	margins := 3 // breathing room
-
-	bodyHeight := a.height - headerLines - tabLines - footerLines - margins
-	if bodyHeight < 5 {
-		bodyHeight = 5
-	}
+	bodyHeight := a.bodyHeight()
 
 	// Constrain body to exact height
 	bodyStyled := lipgloss.NewStyle().
@@ -206,8 +200,8 @@ func (a *App) handleKey(msg tea.KeyMsg) tea.Cmd {
 	case "esc":
 		if inChat {
 			a.chat.Blur()
-			return nil
 		}
+		return nil
 
 	case "q":
 		if !inChat {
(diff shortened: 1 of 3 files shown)


srvsngh99 (commented):
## Code Review

The changes are well-scoped, the code compiles cleanly, and tests pass. A few items worth addressing:

---

### 1. Duplicated layout calculation — extract a helper (medium)

`app.go:150-161` (in `View()`) and `app.go:329-339` (in `resizeViews()`) now contain identical logic for computing `bodyHeight`. This is a maintenance hazard — if either is changed without the other, the bug returns.

**Suggestion:** Extract a `func (a *App) bodyHeight() int` method and call it from both `View()` and `resizeViews()`.

### 2. `margins = 3` is a magic number (low)

Both `View()` (line 159) and `resizeViews()` (line 337) use `margins := 3` with a comment saying "breathing room." Consider a package-level constant:

```go
const layoutMargins = 3
```

### 3. Esc with no fallthrough when not in chat (nitpick)

```go
case "esc":
    if inChat {
        a.chat.Blur()
        return nil
    }
```

When `inChat` is false, Esc falls through to `updateActiveView()` at line 278. Adding an explicit `return nil` for the non-chat case would make intent clearer and prevent accidental forwarding if a future view adds Esc handling.

### 4. Test coverage gap

The new key-handling behavior (Tab always switches, Esc blurs, Shift+Tab always switches) has no unit test coverage. Consider adding tests that:
- Send a `tea.KeyMsg` with Tab when `inChat == true` and assert `activeTab` changed
- Send Esc when focused and assert `Focused() == false`

---

**Overall:** Looks good — approve with the suggestion to extract the duplicated body-height calculation into a shared helper. Test coverage gap is worth a follow-up.

srvsngh99 (commented):
## Re-review: All feedback addressed

All four items from the previous review are resolved:

1. **Shared `bodyHeight()` helper** — Extracted cleanly. `View()` and `resizeViews()` both call `a.bodyHeight()` now. Single source of truth.
2. **`layoutMargins` constant** — Magic number replaced with a named package-level constant with a doc comment.
3. **Esc fallthrough** — Now returns `nil` unconditionally after the `if inChat` block, preventing unintended forwarding.
4. **Test coverage** — Four new tests added covering Tab-always-switches, Shift+Tab-always-switches, Esc-blurs, and Esc-no-forward-on-other-tabs. All passing.

Build, vet, and tests all clean. LGTM.

PR #19, opened 2026-05-01
Fix TUI layout, model pull, and LLM status reporting

srvsngh99:
## Summary

- **Fix init model pull silently dying**: the background goroutine used `cmd.Context()` which was cancelled on process exit, so the download never completed. Replaced with blocking `PullWithProgress()` that shows live download percentage (`Pulling gemma3:4b... 47% (1.4 GB / 3.0 GB)`).
- **Fix dashboard clipping**: use compact header on all tabs (frees ~13 lines), remove always-empty "Active Sub-Krills" panel — the "Did You Know?" fact is now visible.
- **Fix left/right arrow tab switching**: was getting trapped at Chat tab because `onTabSwitch()` auto-focused the input, blocking further arrow navigation. Now matches Tab/Shift+Tab behavior.
- **Fix LLM showing [LIVE] when model not downloaded**: `Available()` only checked if the Ollama *server* was reachable, not if the configured model existed. Now parses `/api/tags` response and checks model presence with name normalization.

## Test plan

- [x] `go build ./cmd/minikrill` compiles clean
- [x] `go test ./...` all pass (19 new test cases for Available() + tab switching)
- [ ] Run `minikrill init`, select Ollama + unpulled model, verify progress bar shows and download completes
- [ ] Run `minikrill tui`, verify compact header, dashboard shows System Status + Did You Know? with fact visible
- [ ] Press left/right arrows across all tabs — no getting stuck at Chat
- [ ] With Ollama running but model not pulled, verify dashboard shows LLM `[DOWN]` not `[LIVE]`

### DIFF
--- a/cmd/minikrill/commands.go
+++ b/cmd/minikrill/commands.go
@@ -135,15 +135,15 @@ var initCmd = &cobra.Command{
 			cfg.Ollama.DefaultModel = selectedModel
 
 			if !localSet[selectedModel] {
-				fmt.Printf(cDim+"  Pulling %s in background..."+cReset+"\n", selectedModel)
-				fmt.Println(cDim + "  Run " + cReset + "minikrill doctor" + cDim + " to check pull status." + cReset)
-				ctx := cmd.Context()
-				go func(model string) {
-					if err := mgr.EnsureRunning(ctx); err != nil {
-						return
-					}
-					_ = mgr.Pull(ctx, model)
-				}(selectedModel)
+				fmt.Println()
+				pullCtx := context.Background()
+				if err := mgr.EnsureRunning(pullCtx); err != nil {
+					fmt.Printf(cYellow+"  Could not start Ollama: %v\n"+cReset, err)
+					fmt.Println(cDim + "  You can pull the model later with: ollama pull " + selectedModel + cReset)
+				} else if err := mgr.PullWithProgress(pullCtx, selectedModel, os.Stdout); err != nil {
+					fmt.Printf(cYellow+"  Pull failed: %v\n"+cReset, err)
+					fmt.Println(cDim + "  You can retry with: ollama pull " + selectedModel + cReset)
+				}
 			}
 		case 1: // Codex
 			cfg.LLM.Provider = "codex"
--- a/cmd/minikrill/commands.go
+++ b/cmd/minikrill/commands.go
@@ -136,14 +136,20 @@ var initCmd = &cobra.Command{
 
 			if !localSet[selectedModel] {
 				fmt.Println()
-				pullCtx := context.Background()
+				pullCtx, pullStop := signal.NotifyContext(context.Background(), os.Interrupt)
+				defer pullStop()
 				if err := mgr.EnsureRunning(pullCtx); err != nil {
 					fmt.Printf(cYellow+"  Could not start Ollama: %v\n"+cReset, err)
 					fmt.Println(cDim + "  You can pull the model later with: ollama pull " + selectedModel + cReset)
 				} else if err := mgr.PullWithProgress(pullCtx, selectedModel, os.Stdout); err != nil {
-					fmt.Printf(cYellow+"  Pull failed: %v\n"+cReset, err)
+					if pullCtx.Err() != nil {
+						fmt.Println(cDim + "\n  Download interrupted." + cReset)
+					} else {
+						fmt.Printf(cYellow+"  Pull failed: %v\n"+cReset, err)
+					}
 					fmt.Println(cDim + "  You can retry with: ollama pull " + selectedModel + cReset)
 				}
+				pullStop()
 			}
 		case 1: // Codex
 			cfg.LLM.Provider = "codex"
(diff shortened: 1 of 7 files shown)


srvsngh99:
## PR #19 Review: Fix TUI layout, model pull, and LLM status reporting

### Positives

- **Root cause fix for the init pull bug** is clean — replacing a fire-and-forget goroutine with a blocking call + `context.Background()` is the right call for a setup wizard.
- **`PullWithProgress` UX** is well done — GB/MB auto-switch, `\r` carriage returns for in-place updates, error propagation from the streaming response.
- **`normalizeModelName`** correctly handles the `gemma3` vs `gemma3:latest` mismatch with Ollama's API. Good edge case.
- **Test coverage** is solid — 7 test cases for `Available()` covering model present/missing/empty/server-down/500/normalization, plus table-driven tests for `normalizeModelName` and 2 new TUI navigation tests.

### Issues to address

**1. Duplicate normalization logic** (minor)

`normalizeModelName` in `internal/llm/ollama.go:280` and the inline `normalize` closure in `internal/ollama/manager.go:482` are identical. One should call the other, or the helper should live in a shared location, to avoid drift.

**2. Left/right arrows now steal input from chat** (medium — please fix)

The old code guarded left/right with `if !inChat` — the PR removes this entirely. This means when a user is typing in the chat input and presses the left arrow to move the cursor within their text, it will switch tabs instead. This is a UX regression. `Tab`/`Shift+Tab` already work unconditionally for tab switching, so left/right should probably remain guarded when the chat input is focused. The PR description says the old behavior "auto-focused the input, blocking further arrow navigation" — but the real fix should be in `onTabSwitch()` not auto-focusing, not in removing the guard entirely.

**3. `PullWithProgress` largely duplicates `Pull`** (minor)

~60% of `PullWithProgress` (`internal/ollama/manager.go:314-396`) is copy-pasted from `Pull` (lines 235-310) — the HTTP request setup, scanner config, NDJSON parsing, error handling. Consider extracting the shared pull-stream logic into a helper that takes a callback for progress handling, to avoid maintaining two nearly-identical streaming parsers.

**4. No timeout or cancellation path for the blocking pull** (low)

The init wizard now blocks on `PullWithProgress` with `context.Background()` — a multi-GB model download with no way for the user to Ctrl+C out (the cobra command's context is not wired in). If Ollama stalls or the network drops, the user is stuck. Consider wiring up signal handling or at minimum using a context that respects SIGINT.

**5. Success line has hardcoded whitespace padding** (nit)

`internal/ollama/manager.go:393`: the trailing spaces in the success message are a fragile way to clear the progress line. If the model name is long, the old progress text could still show through. A `\033[2K\r` (ANSI erase line) would be more robust, or pad dynamically.

### Verdict

The core fixes are sound and well-tested. The **left/right arrow regression in chat** (issue #2) is the main blocker — it breaks text cursor movement while typing. I'd recommend keeping the `!inChat` guard for left/right and instead fixing the root cause (auto-focus in `onTabSwitch` stealing arrow events). The other issues are minor cleanup items that can be addressed in follow-up.

🤖 Reviewed with [Claude Code](https://claude.com/claude-code)

srvsngh99:
All 5 review items addressed in 9a62c21:

1. **Duplicate normalization** — Consolidated into exported `ollama.NormalizeModelName()`. Both `llm/ollama.go` and `ollama/manager.go` now call it from the single source.

2. **Left/right arrow regression** — Restored the `!inChat` guard so cursor movement works while typing. Fixed the root cause instead: `onTabSwitch()` no longer auto-focuses the chat input, so left/right arrows work for tab navigation when arriving at Chat. Press Enter to focus the input and start typing.

3. **PullWithProgress duplication** — Extracted `pullStream()` helper that takes an `onProgress` callback. `Pull()` and `PullWithProgress()` are now thin wrappers (~10 lines each) around the shared streaming logic.

4. **No Ctrl+C for blocking pull** — Wired up `signal.NotifyContext` so the pull respects SIGINT. Shows clean "Download interrupted." message on Ctrl+C.

5. **Hardcoded whitespace** — Replaced trailing spaces with ANSI erase-line (`\033[2K\r`) for robust progress line clearing regardless of model name length.

srvsngh99:
## Re-review after 9a62c21 — All issues addressed ✓

Checked the follow-up commit against all five review items:

| # | Issue | Status |
|---|-------|--------|
| 1 | Duplicate normalization logic | **Fixed** — Exported `ollama.NormalizeModelName`, used by both `llm/ollama.go` and `ollama/manager.go:HasModel`. |
| 2 | Left/right arrows steal chat input | **Fixed** — `!inChat` guard restored on left/right. Root cause addressed: `onTabSwitch()` now always blurs chat instead of auto-focusing, so arrows work for tab nav when you arrive at Chat. Enter re-focuses the input. New tests (`TestLeftRightGuardedWhenChatFocused`, `TestLeftRightSwitchTabsWhenChatBlurred`, `TestEnterFocusesChatWhenBlurred`) cover both paths. |
| 3 | PullWithProgress duplicates Pull | **Fixed** — Shared `pullStream` helper with `onProgress` callback. Clean separation of concerns. |
| 4 | No cancellation for blocking pull | **Fixed** — `signal.NotifyContext` wired up for `os.Interrupt`, with distinct "Download interrupted" message vs other pull errors. `pullStop()` called after the block to clean up. |
| 5 | Hardcoded whitespace padding | **Fixed** — `\033[2K\r` (ANSI erase line) used throughout progress output. |

All tests pass (llm, tui, ollama packages). Build compiles clean.

**LGTM** — this is ready to merge.

🤖 Reviewed with [Claude Code](https://claude.com/claude-code)

PR #20, opened 2026-05-01
Add skill registry TUI tab with enable/disable toggle and default skills

srvsngh99:
## Summary

- **Fix plan bleed bug**: Self-skill matches (e.g. "what can you do") now short-circuit to CHAT intent before LLM classification, preventing them from generating a DIVE PLAN in the chat view
- **Skills tab (tab 3)**: New TUI tab displaying all skills and MCP servers in categorized groups (Built-in, Self-Awareness, Custom/YAML, MCP Servers) with cursor navigation and detail panel
- **Enable/disable toggle**: Press Enter/Space to toggle skills and MCP servers on/off at runtime. Self-awareness skills (`self:*`) are locked and cannot be disabled. Disabled skills are excluded from agent invocations, LLM context injection, and plan generation
- **Default YAML skills**: Ships 4 prompt-template skills out of the box — `summarize`, `explain-code`, `debug`, `brainstorm` — embedded via `go:embed` and seeded to `~/.mini-krill/skills/` on first run without overwriting user edits

## Files changed (16)

| Area | Files |
|------|-------|
| Core interfaces | `internal/core/types.go` |
| Registries | `internal/plugin/registry.go`, `internal/plugin/mcp.go` |
| Default skills | `internal/plugin/defaults.go`, `internal/plugin/defaults/*.yaml` (4 files) |
| TUI | `internal/tui/views.go`, `internal/tui/app.go` |
| Agent | `internal/agent/agent.go` |
| Wiring | `cmd/minikrill/commands.go`, `cmd/minikrill/main.go` |
| Tests | `internal/plugin/registry_test.go`, `internal/agent/agent_test.go`, `internal/tui/app_test.go` |

## Test plan

- [x] `go build ./...` compiles cleanly
- [x] `go test ./...` — all existing + 4 new enable/disable tests pass
- [ ] Launch TUI (`go run ./cmd/minikrill tui`), navigate to Skills tab (press `3`)
- [ ] Verify 4 default skills appear under "Custom / YAML"
- [ ] Toggle a built-in skill off with Enter, verify `[off]` badge
- [ ] Verify `self:*` skills show `locked` and cannot be toggled
- [ ] Switch to Chat tab, verify disabled skill excluded from capabilities
- [ ] Delete a default skill YAML, restart — verify it gets re-seeded


srvsngh99 (commented):
## PR Review: Add skill registry TUI tab with enable/disable toggle and default skills

### Overall
Well-structured PR covering four related features. The code is clean, thread-safe, and has good test coverage. A few design concerns worth addressing before merge.

---

### What looks good

- **Plan bleed fix** is simple and targeted — short-circuiting self-skill matches before the LLM classifier is the right approach.
- **Thread safety** — proper `RWMutex` usage across both registries for enable/disable state.
- **Self-skill locking** — preventing `self:*` skills from being disabled is a nice guardrail.
- **`go:embed` for defaults** — the right choice for bundling YAML skills. `SeedDefaultSkills` correctly avoids overwriting user edits.
- **Test coverage** — 4 new tests covering disable/get, re-enable, self-skill protection, and unknown skill error. All edge cases are hit.
- **TUI** — clean separation of `buildItems()` and `refreshContent()`, cursor navigation is smooth (j/k/g/G), and the detail panel is a nice touch.

---

### Issues to address

#### 1. `Get()` now conflates "disabled" with "not registered" (medium)
**Files:** `registry.go:77-84`, `mcp.go:453-460`

Both `Get()` methods now return `nil, false` for disabled entries — the same return value as "not registered." Callers (e.g. `agent.go:340`, `agent.go:401`) silently skip disabled skills with no log or feedback, which is the desired behavior for runtime filtering. **However**, if any future code needs to distinguish "exists but disabled" from "doesn't exist at all," this API can't do it.

Consider either:
- Documenting this explicitly in the interface contract (acceptable for now), or
- Adding a separate `GetIfEnabled()` method and keeping `Get()` as existence-only (cleaner long-term)

#### 2. Hardcoded `builtinNames` map in the view layer (low-medium)
**File:** `views.go:791-794`

```go
builtinNames := map[string]bool{
    "recall": true, "sysinfo": true, "time": true, "search": true,
}
```

This duplicates knowledge of which skills are built-in. If `RegisterBuiltins()` adds a new skill, this map must be updated manually or the new skill will appear under "Custom / YAML." Consider exposing a `Source` or `Category` field from `SkillInfo` so the registry is the single source of truth.

#### 3. `SeedDefaultSkills` silently swallows `MkdirAll` error (low)
**File:** `defaults.go:47`

```go
_ = os.MkdirAll(dir, 0755)
```

If the directory can't be created (permissions, read-only fs), all subsequent `WriteFile` calls will fail and get logged individually. Better to check the error and return early:
```go
if err := os.MkdirAll(dir, 0755); err != nil {
    log.Warn("could not create skills directory", "path", dir, "error", err)
    return
}
```

#### 4. Enable/disable state is ephemeral (low — just needs documentation)
Toggling a skill off is purely in-memory. Restarting the app resets everything to enabled. If this is intentional, a brief comment or TUI hint like `"(changes reset on restart)"` would prevent user confusion.

#### 5. Toggle error silently dropped in TUI handler (low)
**File:** `views.go:756-762`

```go
_ = s.mcpReg.SetEnabled(item.name, newEnabled)
```

If `SetEnabled` fails (e.g., unknown name due to a race), the user gets no feedback. Since `refreshContent()` re-reads registry state, the UI stays correct, but a flash message or log would help debuggability.

---

### Nits

- `views.go:899-903`: The inner loop to count `catCount` iterates the full items slice for every category boundary. For a small list this is fine, but you could accumulate counts in `buildItems()` and pass them through.
- Default YAML skills: consider adding a `# Auto-generated by mini-krill — edits will be preserved` header to each file so users know they can safely modify them.

---

### Verdict

**Approve with minor changes.** The main concern is #1 (`Get()` semantics) — worth at least a doc comment clarifying the contract. The rest are nice-to-haves that can be addressed in follow-ups.

srvsngh99:
## Re-review: All feedback addressed — LGTM

Commit `90da606` addresses every point from the initial review. Build compiles clean, all tests pass.

---

### Feedback resolution

| # | Issue | Status |
|---|-------|--------|
| 1 | `Get()` conflates "disabled" with "not registered" | **Fixed** — doc comment added to the `SkillRegistry` interface explaining the contract and pointing callers to `IsEnabled` |
| 2 | Hardcoded `builtinNames` map in view layer | **Fixed** — `Category` field added to `SkillInfo`, set by `RegisterBuiltins()`/`RegisterSelfSkills()` via `r.categories` map. View now reads from registry — single source of truth |
| 3 | `SeedDefaultSkills` swallows `MkdirAll` error | **Fixed** — now checks error and returns early with `log.Warn` |
| 4 | Ephemeral enable/disable state undocumented | **Fixed** — TUI footer now shows `"(changes reset on restart)"` |
| 5 | Toggle error silently dropped in TUI | **Fixed** — errors now logged via `log.Debug` |
| nit | `catCount` inner loop per category | **Fixed** — pre-computed in `s.catCounts` map during `buildItems()` |
| nit | Default YAML skills missing user-facing header | **Fixed** — all 4 YAML files now have `# Auto-generated by Mini Krill — your edits will be preserved across updates.` |

### Verification

- `go build ./...` — clean
- `go test ./...` — all pass (agent, plugin, tui, etc.)

No remaining concerns. Ship it.

PR #21, opened 2026-05-01
Promote CLI features to chat skills with TLDR capabilities

srvsngh99:
## Summary

- **5 new chat skills** wrapping CLI-only features so they work from TUI/Chat:
  - `youtube` — paste a YouTube link, get a transcript or AI summary
  - `remind` — natural language reminders ("remind me in 30 min to check the build")
  - `research` — deep web research with sources and citations
  - `web` — read and summarize any web page
  - `notify` — send Telegram notifications from chat
- **TLDR in "what can you do"** — `self:skills` now leads with a plain-language capabilities summary (`=== WHAT I CAN DO ===`) before the detailed skill registry list
- **Conditional registration** — `remind` only registers when a reminder store is available; `notify` only when Telegram is configured

## How it works

All 5 skills reuse existing Go implementations (`internal/content/`, `internal/reminder/`) via thin wrappers. Skills needing state (reminder store, config) receive it through `FeatureContext` at registration time using the same closure pattern as self-skills.

## Files changed (5)

| File | Change |
|------|--------|
| `internal/plugin/features.go` | NEW — 5 feature skills with `FeatureContext` dependency injection |
| `internal/plugin/builtins.go` | `RegisterFeatureSkills()` method |
| `internal/plugin/self.go` | TLDR block prepended to `self:skills` output |
| `cmd/minikrill/main.go` | Wire `RegisterFeatureSkills()` in `initStack()` |
| `internal/plugin/registry_test.go` | `TestFeatureSkillsRegistration` — conditional registration |

## Test plan

- [x] `go build ./...` compiles cleanly
- [x] `go test ./...` — all tests pass
- [ ] TUI: type "what can you do" → verify TLDR appears first with all capabilities
- [ ] TUI: paste a YouTube URL → verify transcript/summary
- [ ] TUI: type "remind me in 5 minutes to stretch" → verify reminder created
- [ ] TUI: type "research best Go web frameworks" → verify research with sources
- [ ] TUI: type "read https://example.com" → verify page summary
- [ ] Skills tab: verify youtube, remind, research, web appear under "Built-in"

### DIFF
--- a/cmd/minikrill/main.go
+++ b/cmd/minikrill/main.go
@@ -166,11 +166,18 @@ func initStack(quiet bool) (*krillStack, error) {
 		DataDir:   config.DataDir(),
 	})
 
+	// Register feature skills (YouTube, reminders, web, research, notify)
+	reminderStore, _ := reminder.NewStore(filepath.Join(config.DataDir(), "reminders.jsonl"))
+	skillReg.RegisterFeatureSkills(plugin.FeatureContext{
+		Config:    cfg,
+		LLM:       providerMgr,
+		Reminders: reminderStore,
+	})
+
 	// Pass recovery config from brain to agent for cold-start continuity
 	cfg.Agent.RecoveryTurns = cfg.Brain.RecoveryTurns
 
 	krillAgent := agent.New(cfg.Agent, providerMgr, krillBrain, skillReg, mcpReg)
-	reminderStore, _ := reminder.NewStore(filepath.Join(config.DataDir(), "reminders.jsonl"))
 	chatHandler := chat.NewHandler(krillAgent, reminderStore)
 
 	return &krillStack{
--- a/cmd/minikrill/main.go
+++ b/cmd/minikrill/main.go
@@ -167,10 +167,12 @@ func initStack(quiet bool) (*krillStack, error) {
 	})
 
 	// Register feature skills (YouTube, reminders, web, research, notify)
-	reminderStore, _ := reminder.NewStore(filepath.Join(config.DataDir(), "reminders.jsonl"))
+	reminderStore, err := reminder.NewStore(filepath.Join(config.DataDir(), "reminders.jsonl"))
+	if err != nil {
+		klog.Warn("reminder store init failed, remind skill will not be available", "error", err)
+	}
 	skillReg.RegisterFeatureSkills(plugin.FeatureContext{
 		Config:    cfg,
-		LLM:       providerMgr,
 		Reminders: reminderStore,
 	})
 
--- a/internal/plugin/builtins.go
+++ b/internal/plugin/builtins.go
@@ -52,7 +52,7 @@ func (r *SkillRegistryImpl) RegisterFeatureSkills(fc FeatureContext) {
 		if err := r.Register(s); err != nil {
 			log.Warn("failed to register feature skill", "name", s.Name(), "error", err)
 		} else {
-			r.categories[s.Name()] = "Built-in"
+			r.categories[s.Name()] = "Feature"
 		}
 	}
 	log.Debug("feature skills registered", "count", len(features))
--- a/internal/plugin/builtins.go
+++ b/internal/plugin/builtins.go
@@ -44,6 +44,20 @@ func (r *SkillRegistryImpl) RegisterSelfSkills(sc SelfContext) {
 	log.Debug("self-awareness skills registered", "count", len(selfSkills))
 }
 
+// RegisterFeatureSkills registers chat skills wrapping CLI-only features
+// (YouTube, reminders, web research, notifications).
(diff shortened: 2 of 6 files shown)


srvsngh99:
## PR Review: Promote CLI features to chat skills and add TLDR

**Commit:** `be5eda0` — 1 commit, 5 files changed, +408/-2

Build compiles, `TestFeatureSkillsRegistration` passes.

---

### What's good

- **Clean dependency gating** — `remind` and `notify` skills are only registered when their dependencies are available (`features.go:38-44`). The test explicitly verifies this.
- **Reuses existing infra** — Skills delegate to `content.ReadYouTube`, `content.Research`, `content.ReadURL`, etc. No logic duplication.
- **Error handling is user-friendly** — Every failure path returns a helpful message with an example, never a raw panic or empty string.
- **Good test** — `TestFeatureSkillsRegistration` covers the conditional registration logic.

---

### Issues to fix before merge

**1. `reminderStore` error silently discarded** (`main.go:170`)
```go
reminderStore, _ := reminder.NewStore(filepath.Join(config.DataDir(), "reminders.jsonl"))
```
The `_` swallows a store initialization error. If `NewStore` fails, `reminderStore` is nil and reminders silently won't register. At minimum add a `log.Warn`.

**2. TLDR is hardcoded and can drift from actual state** (`self.go:242-251`)
The capabilities list ("Summarize YouTube videos", "Set reminders", "Send Telegram notifications") is a static string that doesn't reflect what's actually registered. If Telegram isn't configured, the TLDR still advertises "Send Telegram notifications". Same for reminders if the store fails. Consider filtering this list based on registered skills or adding a caveat like "(when configured)".

---

### Issues — can be follow-ups

**3. `web` skill redirects instead of delegating** (`features.go:146-147`)
```go
if content.IsYouTubeURL(url) {
    return "This looks like a YouTube URL. Try: youtube " + url, nil
}
```
Tells the user to manually re-invoke. Since both skills are in the same package, it could call the YouTube logic directly.

**4. Feature skills categorized as "Built-in"** (`builtins.go:55`)
Both original builtins (recall, sysinfo, time, search) and the new feature skills share `"Built-in"`. A distinct category like `"Feature"` would make the skill list more informative.

**5. `notify` registers but can always fail at runtime** (`features.go:312-314`)
Skill is registered if Telegram **token** exists, but `KRILL_TELEGRAM_CHAT_ID` is only checked at execution time. Consider checking chat ID at registration too.

**6. No test coverage for individual skill execution**
The test verifies registration but none of the skill `Execute` methods are tested — URL parsing, remind subcommand routing, notify validation, etc.

---

### Nits

- `features.go:60` — `urlPattern` is package-level with a very generic name. Fine scoped to these skills, but could be confusing.
- `features.go:99` — Truncation at byte offset 3000 could split a multi-byte UTF-8 character.
- `FeatureContext.LLM` field is set in `main.go:173` but never used — skills receive LLM via `Execute(ctx, input, llm)`. Dead field.

---

### Verdict

Solid feature promotion. **Fix issues #1 and #2 before merge**; the rest can be follow-ups.

srvsngh99:
## Re-review after 184af59

Build passes, tests pass. Checking the two must-fix issues from the previous review:

### Issue #1: reminderStore error silently discarded — FIXED

```go
reminderStore, err := reminder.NewStore(...)
if err != nil {
    klog.Warn("reminder store init failed, remind skill will not be available", "error", err)
}
```
Error is now logged and the nil store gracefully skips remind registration. Good.

### Issue #2: TLDR hardcoded and can drift — FIXED

The TLDR is now driven by `capabilityLines` (`self.go:235-251`) — a table of `(skill name, line)` pairs. At render time it checks which skills are actually registered and enabled, only showing matching lines. Skills with `""` as the skill name are always shown (general capabilities). Clean approach.

### Bonus fixes also addressed

- **Dead `FeatureContext.LLM` field removed** — was a nit, now cleaned up.
- **Feature skills get their own `"Feature"` category** (was issue #4) — `builtins.go:55`, test updated, TUI `views.go` groups them separately.
- **`notify` now checks both token AND chat ID at registration** (was issue #5) — `telegramAvailable` requires both before registering.
- **`web` skill delegates to YouTube directly** instead of bouncing the user (was issue #3).
- **Rune-safe truncation** for transcripts and web content (was a nit).
- **`urlPattern` renamed to `featureURLPattern`** for clarity (was a nit).

### One minor note

`features.go:57-58` has a stale comment on the line before the renamed variable:
```go
// urlPattern is a simple regex to extract a URL from user input.
// featureURLPattern extracts the first URL from user input for feature skills.
```
The first line still says `urlPattern`. Cosmetic only.

### Verdict

Both must-fix issues are resolved. All previously noted follow-up items were also addressed in this commit. **LGTM** — good to merge.

PR #22, opened 2026-05-01
Fix Telegram polling conflict in chat/tui commands

srvsngh99:
## Summary
- `chat` and `tui` commands both called `startBots()` which starts a Telegram `getUpdates` poller — when a `dive` daemon was already running, this created two pollers on the same bot token causing a Telegram API conflict: `"Conflict: terminated by other getUpdates request"`
- Extracted `daemonRunning()` helper from `diveDaemon()` to check if a dive daemon is already active via the PID file
- `chat` and `tui` now skip `startBots()` when a daemon is already running, avoiding the duplicate poller

## Test plan
- [ ] Run `minikrill dive` to start the daemon, then `minikrill chat` — verify no Telegram conflict error
- [ ] Run `minikrill chat` standalone (no daemon) — verify Telegram bot still starts normally
- [ ] Run `minikrill tui` standalone — verify Telegram bot still starts normally
- [ ] Run `minikrill dive` twice — verify it still rejects with "already diving" error

### DIFF
--- a/cmd/minikrill/commands.go
+++ b/cmd/minikrill/commands.go
@@ -367,17 +367,31 @@ var diveCmd = &cobra.Command{
 }
 
 // diveDaemon re-execs the dive command as a detached background process.
+// daemonRunning returns true if a dive daemon is already running.
+func daemonRunning() bool {
+	pidFile := filepath.Join(config.DataDir(), "krill.pid")
+	data, err := os.ReadFile(pidFile)
+	if err != nil {
+		return false
+	}
+	pid, err := strconv.Atoi(strings.TrimSpace(string(data)))
+	if err != nil {
+		return false
+	}
+	proc, err := os.FindProcess(pid)
+	if err != nil {
+		return false
+	}
+	return processRunning(proc)
+}
+
 func diveDaemon() error {
 	// Check if a daemon is already running
-	pidFile := filepath.Join(config.DataDir(), "krill.pid")
-	if data, err := os.ReadFile(pidFile); err == nil {
-		if pid, err := strconv.Atoi(strings.TrimSpace(string(data))); err == nil {
-			if proc, err := os.FindProcess(pid); err == nil {
-				if processRunning(proc) {
-					return fmt.Errorf("Mini Krill is already diving (PID %d). Run 'minikrill surface' first", pid)
-				}
-			}
-		}
+	if daemonRunning() {
+		pidFile := filepath.Join(config.DataDir(), "krill.pid")
+		data, _ := os.ReadFile(pidFile)
+		pid, _ := strconv.Atoi(strings.TrimSpace(string(data)))
+		return fmt.Errorf("Mini Krill is already diving (PID %d). Run 'minikrill surface' first", pid)
 	}
 
 	exe, err := os.Executable()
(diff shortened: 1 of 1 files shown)


srvsngh99 (commented):
## PR Review: Fix Telegram polling conflict in chat/tui commands

### Summary

Good catch on the root cause — two `getUpdates` pollers on the same bot token is a well-known Telegram API conflict. The approach of skipping `startBots()` when a daemon is already running is the right idea.

However, there are a few issues with the implementation that should be addressed before merging.

---

### Issues

**1. TOCTOU race + silent errors in `diveDaemon()` (bug)**

When `daemonRunning()` returns `true`, `diveDaemon()` re-reads the PID file and re-parses the PID just to include it in the error message. Between the two reads, the daemon could be stopped and the PID file removed. The ignored errors (`data, _ := ...` / `pid, _ := ...`) would silently produce `PID 0`:

```
Mini Krill is already diving (PID 0). Run 'minikrill surface' first
```

**Fix:** Have `daemonRunning()` return the PID instead of a bool so `diveDaemon()` doesn't need to re-read the file:

```go
// daemonPID returns the PID of a running dive daemon, or 0 if none.
func daemonPID() int { ... }
```

Then in `diveDaemon()`:
```go
if pid := daemonPID(); pid != 0 {
    return fmt.Errorf("Mini Krill is already diving (PID %d). Run 'minikrill surface' first", pid)
}
```

And callers in `chat`/`tui` become `if daemonPID() == 0 { startBots(...) }`.

**2. Doc comment is merged with the wrong function**

```go
// diveDaemon re-execs the dive command as a detached background process.
// daemonRunning returns true if a dive daemon is already running.
func daemonRunning() bool {
```

The first line is the doc for `diveDaemon`, but it's now attached to `daemonRunning`. These should be separate comment blocks with a blank line between them, or the `diveDaemon` doc should stay above `diveDaemon`.

**3. No user feedback when bots are skipped**

When `chat`/`tui` detect a running daemon and skip `startBots()`, the user gets zero indication. If someone doesn't realize a daemon is running, they'll wonder why their bot config changes aren't taking effect. Consider adding a debug/info log:

```go
if !daemonRunning() {
    startBots(ctx, stack)
} else {
    klog.Debug("skipping bot startup, dive daemon is running")
}
```

---

### Nits

- **`chatCmd` ignores `botStatus`** (pre-existing): `chatCmd` calls `startBots()` without checking the return. `tuiCmd` does check `bs.TelegramErr`. Not introduced by this PR, but worth noting since both paths were touched.
- **`surfaceCmd` duplicates PID-reading logic** (pre-existing): `surfaceCmd` (lines 446–470) has nearly identical PID-file-read-parse-find logic. If you're extracting `daemonRunning()`/`daemonPID()`, this would be a natural place to reuse it too.

---

### Verdict

The fix direction is correct and the scope is well-contained. Issues #1 (race/silent error) and #2 (misplaced doc comment) should be fixed before merge. #3 is a nice-to-have.

srvsngh99 (commented):
## Re-review: All feedback addressed — LGTM ✅

The follow-up commit (`130e3ad`) cleanly addresses all 3 issues and both nits from the previous review:

| # | Item | Status |
|---|------|--------|
| 1 | TOCTOU race / silent errors | **Fixed** — `daemonRunning() bool` → `daemonPID() int`, no more double-read |
| 2 | Misplaced doc comment | **Fixed** — separate doc blocks for `daemonPID` and `diveDaemon` |
| 3 | No log when bots skipped | **Fixed** — `klog.Debug` added in both `chat` and `tui` |
| N1 | `chatCmd` ignoring `botStatus` | **Fixed** — now checks `bs.TelegramErr` like `tuiCmd` does |
| N2 | `surfaceCmd` duplicated PID logic | **Fixed** — reuses `daemonPID()` |

Code is clean, no new issues spotted. Good to merge.

PR #24, opened 2026-05-01
Fix TUI header/tab bar visual duplication

srvsngh99:
## Summary
- Remove the standalone compact header line that rendered in the same bold cyan as the active tab, creating a visual "double tab bar" effect with the 6-tab layout
- Integrate version branding ("Mini Krill v0.1.0") into the tab bar's right-aligned filler area in dimmed style
- Reclaims one vertical row for body content

## Test plan
- [ ] `go build ./...` compiles cleanly, `go test ./...` passes
- [ ] Run `krill tui` — single clean tab bar at the top, no duplicate header row
- [ ] Version branding visible right-aligned in the tab bar area (dimmed)
- [ ] Verify body content has more vertical space than before

### DIFF
--- a/internal/tui/app.go
+++ b/internal/tui/app.go
@@ -8,6 +8,7 @@ import (
 	tea "github.com/charmbracelet/bubbletea"
 	"github.com/charmbracelet/lipgloss"
 
+	"github.com/srvsngh99/mini-krill/internal/brand"
 	"github.com/srvsngh99/mini-krill/internal/core"
 	log "github.com/srvsngh99/mini-krill/internal/log"
 )
@@ -158,8 +159,6 @@ func (a *App) View() string {
 		return "\n  Waiting for terminal..."
 	}
 
-	// Compact header for all tabs to maximize body space.
-	header := RenderCompactHeader(a.version, a.width)
 	tabBar := a.renderTabBar()
 	body := a.renderBody()
 	footer := a.renderFooter()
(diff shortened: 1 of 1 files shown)


srvsngh99 (commented):
## Review: Fix TUI header/tab bar visual duplication

**+16 / -10** | 1 file changed (`internal/tui/app.go`)

### What it does

Removes the standalone `RenderCompactHeader` line that rendered in bold cyan — the same style as the active tab — creating a visual "double tab bar" effect. Version branding (`Mini Krill v0.1.0`) is now right-aligned in the tab bar's filler area using `DimStyle`, reclaiming one vertical row.

### Code verified

- `View()`: removes `header` rendering and its inclusion in `JoinVertical` — straightforward deletion
- `bodyHeight()`: removes `header` from height calculation — keeps layout math consistent
- `renderTabBar()`: adds version tag right-aligned in the filler gap, guarded by `gap > len(versionTag)+4` to avoid overflow on narrow terminals
- `RenderCompactHeader` in `styles.go` is **not deleted** — still used by `DashboardView.View()` for the logo header on tall terminals. No dead code.
- No merge conflicts with the just-merged PR #25.

### Minor nit (non-blocking)

`len(versionTag)` counts bytes, not display width. Fine for ASCII-only `"Mini Krill v0.1.0"`, but `lipgloss.Width(versionTag)` would be safer if `brand.Name` ever includes non-ASCII.

**LGTM. Ready to merge.**

PR #28, opened 2026-05-09
Add smart intent router, real tool execution, and durable tasks

srvsngh99:
## Summary

- **Smart Intent Router** — Replace binary TASK/CHAT classifier with 8-intent router (CHAT, ANSWER, DIAGNOSE, REMEMBER, TOOL_TASK, LONG_TASK, SELF_SKILL, COMMAND). ~80% of inputs handled deterministically with zero LLM calls.
- **Real Tool Execution** — Plan steps now dispatch real tools (shell commands, skills, MCP) instead of LLM narration. Safety-gated: blocked commands, read-only auto-approve, 30s timeout.
- **Durable Task Sessions** — Long-running tasks survive Telegram's 90s timeout via background goroutines with JSONL persistence, async result delivery, and `/tasks` `/task` `/cancel` commands.
- **Smart Plan Approval** — `plan_approval` config changes from bool to 3-way (`auto`/`always`/`never`). Auto mode: ≤4 non-destructive steps execute immediately; 5+ steps or destructive ops require approval.

## Before → After

| Input | Before | After |
|-------|--------|-------|
| "what time is it?" | LLM classify → plan or chat | Deterministic → time skill (0 LLM calls) |
| "why did I get this error?" | TASK → plan → approval | DIAGNOSE → direct answer (1 LLM call) |
| "check the repo" | Plan → narrate file listing | Auto-execute → `ls` + `cat README.md` (real output) |
| "remember I like Go" | Plan about storing | REMEMBER → self:learn skill directly |
| Long task on Telegram | 90s timeout → "context deadline exceeded" | Immediate ack → background execution → async result |
| `rm -rf /` in plan step | LLM might narrate doing it | Blocked outright |

## New files (9)
- `internal/agent/router.go` + `router_test.go` — Intent router
- `internal/agent/toolbox.go` + `toolbox_test.go` — Tool dispatch + shell safety
- `internal/agent/taskstore.go` + `taskstore_test.go` — Durable task persistence
- `internal/agent/taskrunner.go` + `taskrunner_test.go` — Background task execution
- `docs/AUTONOMOUS_AGENT_HANDOFF.md` — Architecture handoff doc (corrected)

## Modified files (8)
- `internal/agent/agent.go` — Router integration, new handlers, task system
- `internal/agent/planner.go` — Tool-aware plan generation and execution
- `internal/core/types.go` — PlanStep.ToolHint, TaskManager interface
- `internal/config/config.go` — PlanApproval bool→string with backward compat
- `internal/chat/telegram.go` — /tasks /task /cancel commands
- `internal/plugin/self.go` — Updated plan approval config handling
- `internal/config/config_test.go`, `internal/agent/agent_test.go` — Updated

## Test plan
- [x] `go test ./...` passes (59 agent tests, all packages green)
- [ ] Start Telegram bot with `minikrill dive --foreground`
- [ ] Send `/help` — verify /tasks, /task, /cancel listed
- [ ] Ask "what time is it?" — verify direct answer, no plan
- [ ] Ask "why did I get this error?" — verify diagnosis, no DIVE PLAN
- [ ] Ask "check the repo" — verify real file reads in output
- [ ] Ask "remember I prefer direct answers" — verify memory stored
- [ ] Send a complex multi-step task — verify background ack on Telegram
- [ ] Send `/tasks` — verify task listing
- [ ] Send `/cancel <id>` — verify cancellation

### DIFF
--- a/cmd/minikrill/commands.go
+++ b/cmd/minikrill/commands.go
@@ -492,16 +492,22 @@ func startBots(ctx context.Context, stack *krillStack) botStatus {
 					AccessedAt: time.Now(),
 				})
 			})
-			// Wire durable task notifications back to Telegram
+			// Wire durable task notifications back to Telegram. RegisterNotifier
+			// scopes this to the "telegram" platform so other platforms (CLI,
+			// TUI, future Discord) don't accidentally receive Telegram-shaped
+			// callbacks.
 			if runner := stack.agent.TaskRunnerRef(); runner != nil {
 				localBot := tgBot // capture for closure
-				runner.SetNotifyFunc(func(platform, chatID, message string) {
-					if platform == "telegram" && chatID != "" {
-						id, err := strconv.ParseInt(chatID, 10, 64)
-						if err == nil {
-							localBot.SendToChat(id, message)
-						}
+				runner.RegisterNotifier("telegram", func(chatID, message string) {
+					if chatID == "" {
+						return
 					}
+					id, err := strconv.ParseInt(chatID, 10, 64)
+					if err != nil {
+						klog.Warn("telegram notify: invalid chat id", "chat_id", chatID, "error", err)
+						return
+					}
+					localBot.SendToChat(id, message)
 				})
 			}
 			go func() {
--- a/cmd/minikrill/commands.go
+++ b/cmd/minikrill/commands.go
@@ -492,6 +492,18 @@ func startBots(ctx context.Context, stack *krillStack) botStatus {
 					AccessedAt: time.Now(),
 				})
 			})
+			// Wire durable task notifications back to Telegram
+			if runner := stack.agent.TaskRunnerRef(); runner != nil {
+				localBot := tgBot // capture for closure
+				runner.SetNotifyFunc(func(platform, chatID, message string) {
+					if platform == "telegram" && chatID != "" {
+						id, err := strconv.ParseInt(chatID, 10, 64)
+						if err == nil {
+							localBot.SendToChat(id, message)
+						}
+					}
+				})
+			}
 			go func() {
 				if err := tgBot.Start(ctx); err != nil {
 					klog.Error("telegram error", "error", err)
--- a/cmd/minikrill/main.go
+++ b/cmd/minikrill/main.go
@@ -179,6 +179,7 @@ func initStack(quiet bool) (*krillStack, error) {
 	cfg.Agent.RecoveryTurns = cfg.Brain.RecoveryTurns
 
 	krillAgent := agent.New(cfg.Agent, providerMgr, krillBrain, skillReg, mcpReg)
+	krillAgent.InitTaskSystem(cfg.Brain.DataDir, 3)
 	chatHandler := chat.NewHandler(krillAgent, reminderStore)
 
 	return &krillStack{
--- a/cmd/minikrill/main.go
+++ b/cmd/minikrill/main.go
@@ -179,7 +179,9 @@ func initStack(quiet bool) (*krillStack, error) {
 	cfg.Agent.RecoveryTurns = cfg.Brain.RecoveryTurns
 
 	krillAgent := agent.New(cfg.Agent, providerMgr, krillBrain, skillReg, mcpReg)
-	krillAgent.InitTaskSystem(cfg.Brain.DataDir, 3)
+	// Tasks are orchestration state, not memories — keep them out of the brain dir
+	// so clearing memory doesn't wipe in-flight task records.
+	krillAgent.InitTaskSystem(filepath.Join(config.DataDir(), "tasks"), 3)
 	chatHandler := chat.NewHandler(krillAgent, reminderStore)
 
 	return &krillStack{
--- a/docs/AUTONOMOUS_AGENT_HANDOFF.md
+++ b/docs/AUTONOMOUS_AGENT_HANDOFF.md
@@ -1,0 +1,353 @@
+# Mini Krill Autonomous Agent Handoff
+
+This document captures the current product gap and a practical execution brief
+for turning Mini Krill from a chat bot with planning into a smarter autonomous
+agent that learns, remembers, and improves with use.
+
+## Goal
+
+Mini Krill should feel like an agent that can:
+
+- understand when a user wants chat, advice, diagnosis, or action
+- inspect local repos and files when asked
+- run real tools and commands with progress updates
+- continue long-running work without Telegram or CLI timeouts killing the task
+- remember durable facts, preferences, projects, and outcomes
+- reflect on repeated feedback and adapt behavior automatically
+
+Today it has pieces of this system, but they are not connected into a real
+agent loop.
+
+## Observed Transcript Symptoms
+
+These behaviors came from real usage:
+
+- Ordinary questions like "why did I get this error?" triggered a full
+  `DIVE PLAN` instead of a direct clarifying answer.
+- Requests like "check the Medha repo in playground" became plans and then
+  failed with `context deadline exceeded`.
+- `/use codex` failed in Telegram even though the README documents `/use`.
+- Approved plans timed out after the user said "yes".
+- The bot sometimes answered with generic personality text instead of useful
+  stateful help.
+- It claimed it could learn and remember, but did not feel like it was evolving
+  from usage.
+
+## Root Causes
+
+### 1. Planner is overused
+
+Relevant file:
+
+- `internal/agent/agent.go`
+
+`classifyIntent` only chooses between `TASK` and `CHAT`. Many messages that
+should be handled as direct answers or lightweight diagnostics become `TASK`,
+which triggers a plan approval loop.
+
+Current issue:
+
+- "Can you check X?" becomes a plan instead of immediate repo inspection.
+- "Why did I get this error?" becomes a plan even when the bot already has
+  enough context to explain likely causes.
+- The user has to approve too many obvious actions.
+
+### 2. Plan execution is only LLM step narration
+
+Relevant file:
+
+- `internal/agent/planner.go`
+
+`ExecutePlanSteps` calls the LLM once per plan step with "Execute this step".
+It does not actually dispatch tools, inspect files, run tests, or persist
+intermediate state.
+
+Current issue:
+
+- A plan can describe useful work without doing useful work.
+- Each step consumes a model call.
+- Multi-step plans regularly exceed Telegram and CLI timeouts.
+
+### 3. Long-running tasks are tied to message timeouts
+
+Relevant files:
+
+- `internal/chat/telegram.go`
+- `cmd/minikrill/cmd_run.go`
+- `internal/llm/cli_provider.go`
+
+Telegram wraps every message in a 90 second context. CLI `run` defaults to 60
+seconds. The CLI provider itself defaults to 10 minutes, but the shorter
+upstream timeouts cancel work before the provider finishes. A multi-step plan
+through Codex or Claude can take longer than both the Telegram and CLI `run`
+limits.
+
+Current issue:
+
+- The user sees `context deadline exceeded`.
+- Work is lost because there is no durable task session.
+- The bot has no way to say "I started this, I will report back."
+
+### 4. Command naming is inconsistent across interfaces
+
+Relevant files:
+
+- `README.md`
+- `docs/INTERFACES.md`
+- `internal/chat/telegram.go`
+- `internal/agent/agent.go`
+
+Both `/use` and `/switch` work, but they are split across interfaces. The core
+agent handles `/use` while Telegram documents and handles `/switch`. Both
+interfaces also support natural language switching ("switch to X", "use X").
+The commands are not broken, but the inconsistent naming confuses users who
+read the README and then try `/use` in Telegram without realizing `/switch` is
+the equivalent there.
+
+Current issue:
+
+- Users may not discover the correct command for their interface on the first try.
+- Documentation should standardize on one name or explicitly list both.
+
+### 5. Memory exists but is not agentic
+
+Relevant files:
+
+- `internal/brain/memory.go`
+- `internal/brain/conversations.go`
+- `internal/agent/agent.go`
+- `internal/plugin/self.go`
+
+Mini Krill stores JSON memories and conversation turns, but memory retrieval is
+mostly substring search and recent-turn replay. Feedback is stored, but
+reflection is manual.
+
+Current issue:
+
+- It remembers raw snippets, not useful knowledge.
+- It does not consolidate repeated facts.
+- It does not automatically learn project context.
+- It does not automatically adapt from feedback.
+
+### 6. Unified conversation can blur context
+
+Relevant file:
+
+- `internal/agent/agent.go`
+
+`SetChannel` stores all turns under one unified channel. This helps continuity,
+but it can blur Telegram group chats, DMs, CLI, and TUI sessions.
+
+Current issue:
+
+- The agent can lose track of who said what.
+- Group memories and private memories need separate scopes.
+
+## Desired Architecture
+
+Build a small autonomous task layer between chat input and LLM/tool execution.
+
+Suggested packages:
+
+- `internal/agent/router.go`
+- `internal/agent/task.go`
+- `internal/agent/toolbox.go`
+- `internal/brain/learning.go`
+- `internal/brain/semantic_memory.go` if embeddings are added later
+
+Core loop:
+
+1. Route intent:
+   - `CHAT`
+   - `ANSWER`
+   - `DIAGNOSE`
+   - `REMEMBER`
+   - `TOOL_TASK`
+   - `LONG_TASK`
+2. For simple chat or answer, respond directly.
+3. For safe obvious tool tasks, execute directly with progress.
+4. For risky or destructive tasks, ask for approval.
+5. For long tasks, create a durable task record and return a task ID.
+6. Execute tools in a worker loop.
+7. Store findings and outcomes in memory.
+8. Periodically reflect and consolidate memories.
+
+## Implementation Plan
+
+### Phase 1: Fix routing and command mismatch
+
+Scope:
+
+- Add a deterministic pre-router before LLM classification.
+- Stop planning for simple questions, greetings, status checks, model switches,
+  and diagnostic questions.
+- Support `/use` and `/switch` consistently in Telegram.
+- Update help text and docs to show both commands or standardize on one.
+
+Acceptance criteria:
+
+- `/use codex` works in Telegram.
+- "why did I get this error?" does not produce a `DIVE PLAN`.
+- "are you okay?" maps to health or chat, not a generic plan.
+- "check repo X" is routed as an actionable tool task, not LLM-only plan text.
+
+### Phase 2: Add durable task sessions
+
+Scope:
+
+- Add a task store under the brain data directory, for example
+  `~/.mini-krill/brain/tasks.jsonl`.
+- Track task ID, status, input, owner/channel, steps, progress, result, error,
+  timestamps.
+- Let Telegram immediately acknowledge long tasks and continue execution in a
+  goroutine.
+- Add commands:
+  - `/tasks`
+  - `/task <id>`
+  - `/cancel <id>`
+
+Acceptance criteria:
+
+- A long repo inspection does not fail because of Telegram's 90 second timeout.
+- The user can ask for task status later.
+- Finished task summaries are saved and recallable.
+
+### Phase 3: Add real local tool execution
+
+Scope:
+
+- Implement a conservative toolbox for local workspace actions:
+  - list files
+  - read files
+  - grep with `rg`
+  - run approved test commands
+  - inspect git status
+- Wire tool execution into task plans.
+- Keep destructive operations blocked unless explicitly approved.
+
+Acceptance criteria:
+
+- "Check the Medha repo in playground" can list the folder, read README/config,
+  identify stack, and summarize next steps.
+- "Go through code and README" performs actual file reads.
+- Task output references real files inspected.
+
+### Phase 4: Improve memory and learning
+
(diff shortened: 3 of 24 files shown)


srvsngh99:
# PR #28 Review — Smart Intent Router, Real Tool Execution, and Durable Tasks

## Overview

Replaces the binary TASK/CHAT classifier with an 8-intent router (deterministic patterns first, LLM fallback for ~20% of cases), wires plan steps to a real `Toolbox` (skills + MCP + sandboxed shell with a 30s cap and a blocked-command list), and adds durable JSONL-persisted background tasks so long jobs survive Telegram's 90s window. `plan_approval` migrates from `bool` to a 3-way string (`auto`/`always`/`never`) with smart heuristics in `auto`. Solid net change: +3214/−233 with 59 tests across 4 new test files. The intent and architecture are sound; the issues below are mostly correctness sharp-edges and one wiring gap.

## Issues / Risks

### 🔴 Likely-broken: legacy bool migration for `plan_approval` (`internal/config/config.go:221`)
`fillDefaults` switches on `strings.ToLower(cfg.Agent.PlanApproval)` and maps `\"true\"`/`\"false\"` to `\"always\"`/`\"never\"`. But `PlanApproval` is now typed `string`, and `yaml.v3` rejects bool→string coercion at unmarshal time. Existing user configs with `plan_approval: true` will fail to load before `fillDefaults` ever runs — the \"backward compat\" path is dead code.
- Fix: implement `UnmarshalYAML` on a wrapper type, accept `interface{}` and normalize, or ship a one-shot migrator. At minimum, add a load test that exercises `plan_approval: true` from a YAML byte slice end-to-end.

### 🟡 TaskRunner concurrency limit has a TOCTOU race (`internal/agent/taskrunner.go:60-67`)
\`\`\`go
current := atomic.LoadInt32(&r.running)
if current >= r.maxConcurrent { return err }
atomic.AddInt32(&r.running, 1)
\`\`\`
Two concurrent `Submit` calls can both pass the check before either increments. Use `CompareAndSwap` in a loop, or `Add(1)` first and roll back if over the limit.

### 🟡 `TaskRunner.notifyFn` data race
`SetNotifyFunc` writes to `r.notifyFn`; the runner goroutine reads it without sync. Either set it once before any Submit (and document that), or guard it with a mutex / `atomic.Value`.

### 🟡 Notify wiring not present in the diff
`SetNotifyFunc` is defined and `taskrunner.notify` is called, but I don't see anywhere in this PR that actually calls `runner.SetNotifyFunc(...)` to bridge it to `TelegramBot.sendMessage`. Background completions will silently no-op on Telegram unless wiring exists outside the diff. Worth confirming and adding a smoke test that exercises Submit→done→notify.

### 🟡 Command-prefix routing over-matches normal text (`internal/agent/router.go: commandPrefixes`)
`/help`, `/status`, `/model`, `/models` are matched with `HasPrefix`, so a message like `/help me debug this` would be routed to `IntentCommand`. Anchor command detection on either an exact match or `prefix + \" \"` (the diff already does this for `/use `, `/auth `, `/task `, `/cancel `; just normalize the rest).

### 🟡 `parseSimpleJSON` is fragile (`internal/agent/planner.go:400-420`)
Splits on `,`, which breaks on values containing commas (e.g., `ARGS: {\"command\": \"ls -la, /tmp\"}`). Tool args from the LLM aren't always trivial — please use `encoding/json` here. The \"we avoid pulling in encoding/json\" rationale doesn't hold (the package is already used in `taskstore.go`).

### 🟡 Auto-mode destructive heuristic false-positives (`agent.go: shouldRequireApproval`)
Substring matching for `\"drop\"`, `\"reset\"`, `\"push\"`, etc. against `step.Description` will trip on benign content like \"drop-down\", \"reset filter state\", \"push notification\". This errs safe (asks for approval) but defeats the auto-execute UX the PR is selling. Consider word-boundary regex, or matching on `ToolHint == \"shell\"` + parsed command rather than free-text.

### 🟢 Minor: `shouldSearch` is now dead code
Comment claims \"other code may still call this\", but with the router taking over, no caller remains in this package. Either delete it or grep-confirm a real caller exists.

### 🟢 Minor: `TaskStore.persistLocked` rewrites the whole file on every update
Acceptable for hundreds of tasks, but you have a JSONL-shaped file used as a snapshot — either rename it (`tasks.json` array) or append-only with periodic compaction. Not blocking.

### 🟢 Minor: `extractShellCommand` heuristics
Falls back to \"everything after `run `/`execute `\", which can produce nonsensical commands from prose like \"run the tests for the auth module\". The `synthesizeToolOutput` path will surface garbage. Consider only honoring the backtick form and routing free-text through the LLM `TOOL_CALL:` path.

## Test Coverage

Strong on router/toolbox/taskstore (~59 tests). Gaps:
- No test exercising the YAML legacy-bool path end-to-end (see top issue).
- No test for the `TaskRunner.Submit` concurrency limit under actual contention (the existing test uses `time.Sleep(20ms)` which is timing-dependent).
- No test that `handleTask` → `submitBackgroundTask` actually fires on `platform == \"telegram\"` with a multi-step plan.

## Security

- Shell sandbox is reasonable: 30s timeout, 10KB output cap, blocked list. The block list is substring-based and works for the obvious patterns.
- Worth noting: `curl <url> | sh` is blocked, but `curl <url> -o /tmp/x && sh /tmp/x` isn't. The block list is a speed-bump, not a sandbox — fine, just don't oversell it.
- `IsReadOnlyCommand` is defined but I don't see it actually gating execution in `ExecuteShellSafe`. Currently *every* non-blocked command runs. If the intent was \"read-only auto-approve, write requires approval\", that gating isn't wired up here — `NeedsApproval` on the step is only set by destructive-keyword scanning of the description, not by inspecting the actual shell command.

## Recommendation

Request changes for the YAML migration (real correctness bug for existing users) and the two concurrency races. The other items are good follow-ups but not blockers. The architectural direction — deterministic routing + real tools + durable tasks — is the right call and the test density is much better than the previous plan-only path.

srvsngh99:
# Re-Review (after `ece76c4` + `05be211`)

Quick follow-up after the fix commit and the new memory-consolidation commit. Most of the original blockers are resolved cleanly; one real concurrency bug remains and there's some new surface to consider.

## ✅ Verified fixed

| Original issue | Fix |
|---|---|
| YAML legacy bool migration | Custom `UnmarshalYAML`/`MarshalYAML` on `AgentConfig` via `interface{}` intermediary; end-to-end tests for `true`/`false`/`auto`/`always`/`never` (`config.go:39-90`, `config_test.go:197-250`). Solid. |
| TaskRunner Submit TOCTOU race | `Add(1)` first then rollback if over the limit (`taskrunner.go:50-59`). Correct. |
| `notifyFn` data race | `sync.RWMutex` guarding `notifyFn` (`taskrunner.go:24-25, 44-48, 110-118`). Correct. |
| Command-prefix over-match | Split into `commandExact` (`/help`, `/model`, `/models`, `/status`, `/tasks`) and `commandPrefixes` (`/use `, `/auth `, `/task `, `/cancel `). New test `TestRouterCommandDoesNotOvermatch`. |
| `parseSimpleJSON` fragility | Now uses `encoding/json.Unmarshal` as the primary path, comma-split kept as fallback. |
| `extractShellCommand` heuristic | Restricted to backtick-quoted commands; "run/execute" prose path removed. Test updated. |
| Destructive heuristic false positives | Switched to specific phrases (`"delete file"`, `"git push"`, `"drop table"`, `"reset --hard"`, etc.); `isWordChar` now treats `-` as a word char so "drop-down" doesn't tokenize as "drop". `TestShouldRequireApprovalAutoNoFalsePositives` covers it. |
| `shouldSearch` dead code | Removed. |

## 🔴 New blocker — `go test -race` fails

`TestTaskRunnerSubmitAndComplete` fails under `-race` on the current branch:
```
WARNING: DATA RACE
Write at 0x... by goroutine 7:
  taskrunner.go:73   r.store.Update(task.ID, "running", "", "")
Previous read at 0x... by goroutine 6:
  taskrunner_test.go:49   if ok && (got.Status == "done" || got.Status == "failed")
```

Root cause is in `TaskStore`, not the test: `Get(id)` and `List(userID)` return `*DurableTask` pointers under only the briefest read-lock, so any caller that reads `.Status` (or any other field) is racing the `Update()` writer that holds `s.mu` while mutating the same struct. The earlier `notifyFn` race fix didn't cover this one.

Two options:
1. Make `Get`/`List` return value copies (`DurableTask`, not `*DurableTask`). Cheapest, fixes everything.
2. Keep returning pointers but require callers to lock — unergonomic; will leak.

Option 1 is what the existing `ToInfo()` / `core.TaskInfo` API already implies elsewhere, so I'd lean that way.

Reproduce:
```
go test -race -run TestTaskRunnerSubmitAndComplete ./internal/agent/...
```

## 🟡 Wiring gap still unaddressed

`SetNotifyFunc` is now race-free, but I still don't see a call site in this PR that actually wires it to `TelegramBot.sendMessage`. Background-task completions will still no-op on Telegram unless wiring exists outside the diff. Consider adding a brief integration smoke test, or at minimum a comment on `InitTaskSystem` pointing at where the caller is expected to register the notifier.

## 🟢 Notes on the new memory-consolidation commit (`05be211`)

Direction is good — scoped memory + ranked retrieval is a real improvement over substring search. Two minor things:

- **`RankedSearch` sort is hand-rolled O(n²)** (`memory.go:181-187`). For a modest store this is fine, but please use `sort.Slice` — same line count, predictable behavior, easier to reason about.
- **`Consolidate` cluster discovery is also O(n²)** with no upper bound on entries scanned. With a long-lived memory store this could become a noticeable wall-clock hit when triggered by `self:consolidate`. Worth a `limit` parameter or a soft cap with a "too many entries to consolidate in one pass" log line.
- **`MemoryConsolidator` only activates if memory is `*brain.FileMemory`** (type assertion in `taskrunner.go:storeAsMemory`). That's pragmatic for now, but it means anyone swapping in a different `core.Memory` implementation silently loses reflections. Consider lifting `ReflectOnTask` to an interface method on `core.Memory` or a separate `core.Reflector` interface so the dependency is explicit.

## Recommendation

The PR is close to mergeable. The only remaining blocker I can reproduce is the `TaskStore.Get`/`List` race — please copy on read. The notify-wiring gap is worth confirming before merge, even if it's actually wired up outside this diff. The memory-consolidation additions are nice-to-have polish; nothing in there is a blocker.

srvsngh99:
# End-to-End Review (after `ece76c4`)

Pulled the branch fresh, built the binary, and exercised the new code paths against a real `KRILL_DATA_DIR`. Build is clean, `go vet ./...` is clean, plain `go test ./...` is green. Most of the previously-flagged correctness issues are fixed. But the headline \"durable task sessions\" feature is **shipped completely dark** end-to-end, and the `go test -race` regression I flagged in the previous re-review is still live.

## 🔴 Blocker — durable tasks are not wired into the binary

`InitTaskSystem`, `SetPlatform`, and `SetNotifyFunc` are defined on `*KrillAgent` but **have zero callers in non-test code**. Verified via `grep -rn 'InitTaskSystem\|SetNotifyFunc\|SetPlatform' --include='*.go'`:
- `cmd/minikrill/main.go:181` constructs the agent with `agent.New(...)` and never calls any of the three.
- `internal/chat/handler.go` logs `msg.Platform` but never propagates it to the agent.
- `internal/chat/telegram.go` — no notifier wiring.

Live trace against the built binary in a fresh data dir:
```
$ minikrill run \"/tasks\"
Task system is not enabled.
$ minikrill run \"/task abc\"
Task system is not enabled.
$ minikrill run \"/cancel abc\"
Task system is not enabled.
```

Downstream consequences (all observable, not theoretical):
- `/tasks`, `/task <id>`, `/cancel <id>` are advertised in `/help` but always return \"Task system is not enabled.\"
- `shouldRunInBackground(plan)` checks `a.taskStore == nil` first → always returns `false`. So `submitBackgroundTask` is unreachable production code, the \"On it! I'll work on this in the background\" ack the PR description promises is never sent, and Telegram's 90s timeout will still bite on long tasks — exactly the symptom this PR claims to fix.
- `IntentLongTask` falls through to synchronous `ExecutePlan`, identical to pre-PR behavior.

Required wiring (rough sketch for `cmd/minikrill/main.go` after `agent.New`):
```go
krillAgent.InitTaskSystem(cfg.DataDir, 3)
if runner := krillAgent.TaskRunnerRef(); runner != nil {
    runner.SetNotifyFunc(func(platform, chatID, message string) {
        // dispatch back to the right chat platform
    })
}
```
Plus per-message `SetPlatform` (or — better — make platform a per-call argument; the current single-field design races between concurrent Telegram + CLI users).

There also needs to be a notifier registry keyed by platform (telegram → `TelegramBot.sendMessage`, discord → ditto, cli/tui → no-op or stdout). One single `NotifyFunc` on a process-wide `TaskRunner` doesn't model multi-platform delivery cleanly — fine for v1 but worth a TODO.

## 🔴 Still failing — `go test -race`

`TestTaskRunnerSubmitAndComplete` still trips the race detector:
```
WARNING: DATA RACE
Write at … by goroutine 7:
  taskrunner.go:73   r.store.Update(task.ID, \"running\", …)
Previous read at … by goroutine 6:
  taskrunner_test.go:49   if ok && (got.Status == \"done\" || got.Status == \"failed\")
```
Root cause is unchanged from my last review: `TaskStore.Get` and `List` return `*DurableTask`, callers read `.Status`/`.Result` after releasing the read lock, while `Update` mutates the same struct under the write lock. Cheapest fix: return value copies (`DurableTask`, not `*DurableTask`); the JSONL persistence already round-trips via copies anyway. Equivalent change to `List` returning `[]DurableTask`. `ToInfo()` callers don't care.

`go test ./...` (no `-race`) passes — but please don't merge a feature whose race-detector signal is permanently red, especially when the feature is concurrency-heavy by definition.

## 🟡 Single-platform field on a process-wide agent

`a.platform` is one mutable string set by `SetPlatform`. Even when you wire it, two concurrent inbound messages (Telegram + CLI, or two Telegram users on different chats) would race the field and `shouldRunInBackground` would consult whichever value won the race. Either:
- Pass `platform` as an argument through `Chat()` and into `handleTask`, or
- Store it on the per-message context, or
- Snapshot it inside the Chat mutex before any branching.

Right now the whole `Chat()` function holds `a.mu`, which serializes all messages globally — so the race doesn't manifest *under that lock*, but the lock itself is a separate scaling problem worth flagging.

## 🟡 `commandExact` blocks the natural-language fallback for `/help me debug`

The router fix correctly distinguishes exact-match commands from prefix commands. Good. But for an exact like `/status`, if a user types `/status` alone they'll hit the `IntentCommand` branch in `Chat()` → `handleTaskCommand` returns `\"\", false` → falls through to `handleChat`. `handleProviderCommand` doesn't handle `/status` either (verified — no case for it). So the user sees a generic chat response to a slash-command they probably expected to be informative. Either add a real handler for `/status` or remove it from `commandExact` so the router doesn't classify it as a command at all.

## 🟢 Confirmed working end-to-end

| Path | Live test | Result |
|---|---|---|
| `IntentToolTask` → time skill | `minikrill run \"what time is it\"` | ✅ Returns real `Current time: 2026-05-10 03:47:22 IST`, no LLM plan |
| Build | `go build ./...` | clean |
| Vet | `go vet ./...` | clean |
| Plain test suite | `go test ./...` | all packages green |
| YAML legacy bool migration | `TestPlanApprovalLegacyBool{True,False}` | passing |
| Router deterministic paths | router_test.go suite | passing |
| Auto-mode false-positive avoidance | `TestShouldRequireApprovalAutoNoFalsePositives` | passing |

## 🟢 Memory consolidation (`05be211`) sanity

Tests pass, scoring math is reasonable. Notes from before stand: O(n²) sort and O(n²) cluster discovery are fine at current scale but worth `sort.Slice` and a soft cap. The `*brain.FileMemory` type assertion in `TaskRunner.storeAsMemory` silently drops reflections for any future `core.Memory` implementation — lift `ReflectOnTask` to an interface.

## 🟢 Other things I checked

- `shouldSearch` removal: confirmed no lingering callers (`grep -rn 'shouldSearch'` returns nothing).
- `parseSimpleJSON` JSON-first parsing: works on well-formed `{\"key\": \"value with, comma\"}`.
- `extractShellCommand`: backtick-only is correct; `Run \`ls\`` extracts, prose like `\"run the tests\"` correctly returns `\"\"`.
- Self-skill order: write triggers checked before read triggers, so `\"evolve your personality\"` correctly maps to `self:evolve` not `self:inspect`.

## Recommendation

**Hold merge.** Two real issues:
1. The durable-task subsystem is dead end-to-end — `/tasks`/`/task`/`/cancel` return error, Telegram timeouts are not actually fixed. The PR description significantly overstates current behavior. Either wire the system in this PR (add ~10 lines to `main.go` and a notifier in `telegram.go`), or pull the user-facing surface (help text, command routing) out of this PR and ship the wiring together with it.
2. `TaskStore.Get`/`List` pointer race fails `go test -race`. Two-line fix.

The architectural work itself is good and the iteration on review feedback has been thorough; just need the last mile of wiring + the copy-on-read fix before this is shippable.

srvsngh99:
# Re-Review (after `469faf36`)

Pulled fresh, rebuilt, exercised the same end-to-end traces from the previous review. Both blockers are resolved.

## ✅ Verified fixed

| Previous blocker | Fix |
|---|---|
| Durable task system not wired | `cmd/minikrill/main.go:182` calls `InitTaskSystem(cfg.Brain.DataDir, 3)`; `commands.go:495-505` wires `TaskRunner.SetNotifyFunc` to `TelegramBot.SendToChat`; `chat/handler.go:53-55` calls `SetChatContext(platform, chatID)` before each `agent.Chat`. Live test: `minikrill run "/tasks"` now returns `"No tasks found."` (was: `"Task system is not enabled."`). `/task abc` returns `"Task \"abc\" not found."`. `/cancel abc` returns proper error. |
| `go test -race` failure | `TaskStore.Get`/`List` now return value copies (`DurableTask`, not `*DurableTask`). `TestTaskStoreCreateAndGet` adds an explicit "mutation should not leak back" assertion. **`go test -race ./...` now passes across all packages — zero races detected.** |
| `/help me debug this` over-match | `/help` and `/status` removed from `commandExact` (handled by platform bot dispatch directly). Live test: `minikrill run "/help me debug this error"` correctly routes to chat/diagnose, not COMMAND. |
| `RankedSearch` O(n²) sort | Replaced with `sort.Slice`. |

## 🟡 Remaining: `SetChatContext` + `Chat` not atomic

The single shared `platform`/`chatID` fields on `*KrillAgent` plus `Chat`'s full-duration mutex make this work for sequential single-bot use, but the public API still has a race window:

```
handler1: SetChatContext("telegram", "123")   // lock-set-unlock
handler2: SetChatContext("cli", "0")          // lock-set-unlock  ← overwrites
handler1: Chat(...)                           // reads platform="cli" — WRONG
```

Two concurrent dispatchers (e.g. Telegram + TUI in the same process) can interleave their `SetChatContext`+`Chat` pairs and end up with the wrong platform attributed to a task. The comment in `agent.go` claims "they cannot race" — true at the field-tearing level but not at the invariant level. The current production deployment (single Telegram bot, one message at a time) won't hit this because the poller dispatches sequentially and `Chat()`'s mutex serialises everything else, but it's a footgun waiting for the first multi-platform user.

Cleanest fix: add `platform`/`chatID` as parameters to `Chat(ctx, input, platform, chatID)`, drop the agent fields, and remove `SetChatContext`. The field is only read in two places (`shouldRunInBackground` and `submitBackgroundTask`), so the change is localised. Not a merge blocker, but worth a follow-up issue.

## 🟢 Minor observations (non-blocking)

- **`tasks.jsonl` lives under `brain/`**: `InitTaskSystem(cfg.Brain.DataDir, ...)` puts the JSONL alongside memory files. Tasks aren't memories — they're orchestration state. Consider a separate `cfg.Agent.DataDir` (or just `cfg.DataDir`) so future operators don't accidentally wipe task state when clearing the brain.
- **`SetNotifyFunc` is single-platform**: only Telegram gets a notifier wired (Discord/CLI/TUI background completions silently drop). The `if platform == "telegram"` gate in the closure is correct for the current setup but the wider design assumes one notifier per process. Fine for v1; revisit when Discord support lands.
- **Type-assertion duck-typing in `handler.go:53`**: `if ps, ok := h.agent.(interface{ SetChatContext(string, string) }); ok` works but means a renamed method silently disables task routing. Either add `SetChatContext` to a `core.Agent`-adjacent interface, or call it directly on `*KrillAgent` with no fallback.
- **`/help me debug this` did go through correctly**, but the LLM response leaked an internal directive token (`[CROSSPOST:chat_id]request_error_paste[/CROSSPOST]`). Prompt-template artifact unrelated to this PR — flagging only because it surfaced during the live test.
- **Skill init order**: the skill registry is built before `InitTaskSystem`, so a skill that wants `core.TaskManager` at registration time can't. Not a problem today, but a future concurrent-task skill would need lazy lookup.

## Test summary on this branch

```
go build ./...        clean
go vet ./...          clean
go test ./...         all green
go test -race ./...   all green ← previously red
```

Live binary smoke trace:
```
$ minikrill run "/tasks"             → "No tasks found."
$ minikrill run "/task abc"          → "Task \"abc\" not found."
$ minikrill run "/cancel abc"        → "Cancel failed: task \"abc\" not found"
$ minikrill run "/help me debug…"    → routed to diagnose path (not COMMAND)
$ minikrill run "what time is it"    → real time skill output, no LLM plan
```

## Recommendation

**Approved to merge.** All previously identified blockers (YAML migration, two concurrency races, command over-match, JSON parsing, shell extraction, destructive heuristic, end-to-end task wiring) are resolved with tests. The remaining `SetChatContext` atomicity issue is a real footgun but won't bite the current single-bot deployment, and the cleanup is small enough for a follow-up. Memory consolidation work (`05be211`) is a clean addition with the O(n²) sort fixed.

Nice iteration cycle on the review feedback — went from architecture-sound-but-half-wired to actually shipping the feature in two commits.

PR #29, opened 2026-05-10
Make agent smart and autonomous instead of rigid

srvsngh99:
## Summary
- Replaces the static yes/no approval gate with a regex fast-path + LLM-judged classifier. Accepts casual approvals ("yea", "ok", "alright"), supports a MODIFY outcome that regenerates the plan, and releases control on UNRELATED replies — fixing the loop where the same prompt re-rendered forever.
- Adds a typed user-preference layer (`internal/agent/prefs.go`) so phrases like "no need for approval", "stop asking me to approve", "be terse", "no metaphors" get captured as stable boolean flags backed by `brain.Memory()`. `shouldRequireApproval` reads the flag under "auto" mode while destructive plans still gate.
- Stops forcing plans onto casual asks: `actionVerbs` is split into hard (deploy/install/refactor/build/fix) vs soft (find/give/suggest/brainstorm). Soft verbs alone keep the message in CHAT/ANSWER. The LLM classifier prompt is rewritten so ideation/brainstorming never lands in LONG_TASK.
- Real auto tool dispatch: broadened `searchTriggers` ("find me", "news on", "digest", "today in") and added a freshness auto-route ("today" + ≥3 words → search). The system prompt now explicitly forbids hallucinating CLI permission/approval prompts for built-in tools (the previous behaviour invented a permission gate that doesn't exist).
- Adaptive personality: `maybeAutoEvolveStyle` flips `pref:style_terse` automatically after 3 corrections within an hour. `handleChat` / `handleAnswer` inject `buildStyleDirective` so active prefs become per-turn system messages.

### Behavioural deltas (mapped to the original transcript)
| Transcript moment | New behaviour |
| --- | --- |
| "Give me some ideas" → DIVE PLAN | Soft-verb only, ≤6 words → CHAT/ANSWER. No plan, no approval. |
| "No need for taking my approval" | `pref:no_plan_approval=true` written; future plans skip the gate. |
| "Yea" rejected | Regex now accepts yea/yeah/yep/yup/y/sure/ok/alright/👍. |
| Loop on ambiguous input | UNRELATED releases the pending plan and re-routes; MODIFY regenerates. |
| "find me some AI digest from today" → fake permission | Routed straight to the search skill via freshness heuristic; soul prompt forbids fabricated permission prompts. |

## Test plan
- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `go test ./...` all packages green
- [x] 22 new cases in `prefs_test.go` (typed prefs, regex coverage, pending-plan flow, auto-evolve, destructive detection)
- [x] 3 new router test groups (soft vs hard verbs, freshness routing)
- [ ] Manual smoke test: trigger the original transcript flow end-to-end to confirm the agent no longer loops on "yea" / "no need for approval"

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
 	"sync"
 	"time"
@@ -30,31 +31,13 @@ const maxHistory = 20
 
 const unifiedConversationChannel = "unified"
 
-// approvalWords are inputs that greenlight a pending plan.
-var approvalWords = map[string]bool{
-	"yes":      true,
-	"approve":  true,
-	"go":       true,
-	"do it":    true,
-	"lgtm":     true,
-	"go ahead": true,
-	"y":        true,
-	"yep":      true,
-	"sure":     true,
-	"proceed":  true,
-}
+// approvalRegex matches obvious approval shorthand on the fast path so we
+// don't burn an LLM call on "yes". Anything not matched here OR by
+// rejectionRegex falls through to the LLM-judged classifier.
+var approvalRegex = regexp.MustCompile(`^(?:y|yes|yea|yeah|yep|yup|ya|sure|ok|okay|k|👍|do it|go|go ahead|let'?s go|lgtm|proceed|approve|approved|fine|alright|sounds good)[\s\.!]*$`)
 
-// rejectionWords are inputs that scrap a pending plan.
-var rejectionWords = map[string]bool{
-	"no":     true,
-	"reject": true,
-	"cancel": true,
-	"stop":   true,
-	"nah":    true,
-	"nope":   true,
-	"n":      true,
-	"abort":  true,
-}
+// rejectionRegex covers obvious "scrap it" shorthand.
+var rejectionRegex = regexp.MustCompile(`^(?:n|no|nah|nope|cancel|cancelled|stop|abort|reject|rejected|skip|nevermind|never mind|forget it|don'?t)[\s\.!]*$`)
 
 // KrillAgent is the main agent that implements core.Agent.
 // It is the executive function of the krill brain - classifying intent,
@@ -228,8 +211,19 @@ func (a *KrillAgent) ChatFromPlatform(ctx context.Context, platform, chatID, inp
 	}
 
 	// --- Phase 1: Check for pending plan approval ---
+	// Even with a pending plan, we still want preference learning to fire so
+	// "stop asking me to approve" gets captured *during* the approval state,
+	// not after. The pending-plan handler itself decides whether to consume
+	// the input or release it back to normal routing.
 	if a.pendingPlan != nil {
-		return a.handlePendingPlan(ctx, input)
+		a.maybeStoreUserPreference(ctx, input)
+		response, handled := a.handlePendingPlan(ctx, input)
+		if handled {
+			return response, nil
+		}
+		// Fell through: input was unrelated to the pending plan. Drop the
+		// pending plan and route the input normally.
+		a.pendingPlan = nil
 	}
 
 	// --- Phase 2: Record user message ---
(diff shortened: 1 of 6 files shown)


srvsngh99:
## Code Review — PR #29: Make agent smart and autonomous instead of rigid

### Overview
Replaces the static yes/no approval gate with a regex fast-path + LLM-judged classifier (4 outcomes: APPROVE / REJECT / MODIFY / UNRELATED), introduces a typed user-preference layer (`internal/agent/prefs.go`) over `brain.Memory()`, splits action verbs into `hard` vs `soft` so ideation no longer falls into LONG_TASK, broadens search auto-routing with a freshness heuristic, and adds an adaptive style flip after repeated corrections. Net result: the original transcript bugs (loop on "yea", fabricated "permission required" responses, ideation routed to plan mode) are addressed with targeted changes plus solid test coverage.

### What's good
- **Clean layering.** `prefs.go` gives decision sites (`shouldRequireApproval`, `buildStyleDirective`) stable boolean keys instead of scanning free-text memory. Set/Get is a thin, well-typed shim over `core.Memory`.
- **Safety preserved under autonomy.** `shouldRequireApproval` checks `planIsDestructive` *before* honouring `pref:no_plan_approval`, and `config.PlanApproval == "always"` still wins. Tested with `TestShouldRequireApproval_PrefDoesNotSkipDestructive` and `..._PrefDoesNotOverrideAlways`.
- **Good test coverage of edge cases.** Regex coverage (incl. emoji 👍, "alright!", "yes but change step 3" intentionally NOT matching), destructive false-positive guard ("drop-down"), threshold boundaries (`maybeAutoEvolveStyle` recent vs >1h old), and the central transcript-regression test (`TestPendingPlan_YeaApproves`).
- **Soul prompt update is precise.** The new rule explicitly forbids fabricated "Allow button" / "CLI permission prompt" responses — the exact failure mode from the transcript.
- **Cleanup commit (81bbcee) already removed the legacy `actionVerbs` / `softActionVerbs` dead code** that I was about to flag.

### Issues / suggestions

**1. Double-counting risk in `maybeAutoEvolveStyle` (agent.go:990-1006).** The comment says the "+1 for the signal we just observed" is a guard against backends that don't return the just-stored entry. But `recordFeedback` *does* store the signal before this function runs (the call chain is `recordFeedback → store → maybeAutoEvolveStyle`). If `Memory.Search` returns it, `count` is already N+1, and the explicit `count++` makes it N+2 — the threshold trips one signal early. Tests don't catch this because they don't pre-store the latest signal.
- Fix: either drop the `count++` and rely on Search returning fresh entries, or pass a `latestStoredKey` and skip it during the loop. Or use a CreatedAt cutoff that excludes the very-most-recent entry.

**2. `shouldRequireApproval` uses `context.Background()` (agent.go:639).** The call site has a live `ctx` available (`handleTask` runs inside the request). Plumbing it through means user cancellation / deadlines apply to the memory `Recall` instead of using a detached background context.
- `shouldRequireApproval(plan)` → `shouldRequireApproval(ctx, plan)`.

**3. No inverse trigger to flip `PrefStyleTerse` back off.** Once auto-evolve flips it, the only way to undo is unwritten ground. `detectTypedPreference` lacks "be verbose", "more detail", "longer responses". Consider adding them so the loop is symmetric — otherwise the "they can undo this by asking for more detail" claim in the PR body isn't actually wired.

**4. LLM-judge `MaxTokens=8` is tight (agent.go:475).** Category names tokenize differently across providers; "UNRELATED" can be 2-3 tokens, plus instruction-following models sometimes emit a leading newline/space. With 8 tokens you may truncate to "UNRELATE" and the `strings.Contains(out, "UNRELATED")` matcher fails → falls through to the `default` case, which IS UNRELATED, so it's mostly self-healing — but the same logic would silently misclassify a truncated "MODI" → falls to UNRELATED instead of MODIFY. Bumping to 16-32 has negligible cost.

**5. `maybeStoreUserPreference` may run twice on UNRELATED path.** In `ChatFromPlatform`: when `pendingPlan != nil` we call it on line 219, then `handlePendingPlan` returns `(_, false)`, we null the plan and fall through to Phase 6 where `maybeStoreUserPreference` is called again on the same input. Typed prefs are idempotent (same key, same value), but free-text memory uses time-stamped keys → two near-duplicate entries per UNRELATED turn. Cheap to fix: add a sentinel/early-return after the pending-plan path stores it.

**6. Approval-regex single-letter trip wires.** `n` and `y` alone match. Realistic but worth confirming that voice-typed / fat-fingered "n" doesn't get treated as rejection in your real chat traffic. The regex is greedy enough that "nope but actually..." stays unmatched (good) — only the bare letter is the issue. Low priority.

**7. Per-turn cost of `buildStyleDirective`.** Three `Memory.Recall` calls per chat/answer turn. For SQLite this is negligible; for any future remote-backed memory, consider caching the resolved directive on the agent struct and invalidating on `SetBoolPref`.

**8. Style directive injection pattern uses `append(slice[:N-1], a, slice[N-1])` twice in succession (agent.go:687-690 and again in handleAnswer).** Works correctly with current capacity, but the aliasing pattern is fragile. Pulling out a small `insertBeforeLast(msgs, newMsgs...)` helper would document the intent and prevent subtle bugs if anyone refactors.

### Risks
- **Behavioral**: A user who once says "no need for approval" stays opted-out forever (no decay). For a personal-assistant agent this is fine; flag it in your own head if multi-user scenarios appear.
- **LLM cost**: Every ambiguous reply during a pending plan now spends an extra LLM call on classification. Bounded — only fires when regex misses — but worth watching in metrics.
- **Auto-terse flip is unidirectional** (see #3) and triggers on 3 corrections/hour, which is a low bar during exploratory sessions.

### Verdict
Targeted fixes for real bugs, not a kitchen-sink rewrite. Ship after addressing #1 (double-counting) and #2 (ctx plumbing); the rest are polish. Good test discipline throughout.

srvsngh99:
Thanks for the thorough review. Pushed [70359b3](https://github.com/srvsngh99/mini-krill/commit/70359b3) addressing the must-fixes plus the high-value polish:

- **#1 (double-count)** — `maybeAutoEvolveStyle` now takes a `latestKey` parameter; the search loop skips that key, the explicit `+1` adds it once. Regression test `TestMaybeAutoEvolveStyle_NoDoubleCountOnLatestKey` covers the case where only 2 distinct signals exist with the latest already stored — without the fix, count would be 3 and trip the threshold; with the fix it stays at 2.
- **#2 (ctx plumbing)** — `shouldRequireApproval(ctx, plan)`, threaded from `handleTask`. Test callsites updated.
- **#3 (inverse trigger)** — `detectTypedPreference` now recognises "be verbose", "more detail", "longer responses", "be more thorough" → `PrefStyleTerse=false`. Test added.
- **#4 (LLM judge MaxTokens)** — bumped 8 → 24 with a comment on why.
- **#5 (duplicate `maybeStoreUserPreference`)** — added a `prefStored` flag in `ChatFromPlatform` so the UNRELATED release path doesn't write twice.
- **#8 (insertBeforeLast helper)** — extracted; replaced all 10 call sites. Always allocates a fresh backing array.

Skipped:
- **#6** — single-letter "n"/"y" trip wires. Low priority; would only matter for fat-fingered single-key chat. Happy to tighten if you see it in real traffic.
- **#7** — caching `buildStyleDirective`. Three `Memory.Recall` per turn against the file-backed store is below noise floor; deferring until a remote-backed memory exists.

Behavioural caveat acknowledgements:
- "No-decay opt-out" — kept as-is for personal-assistant use; will revisit if multi-user shows up.
- "LLM cost on ambiguous replies" — bounded by the regex fast-path; will watch in metrics once instrumented.
- "Auto-terse unidirectional" — addressed by #3, no longer the case.

CI green on the previous push (lint/test/build all pass). Ready for another look.

srvsngh99:
## Follow-up review — 70359b3

Verified all 6 addressed items against the diff. Tests pass locally (`go test -count=1 ./internal/agent/...` clean).

| # | Issue | Status | Verification |
| --- | --- | --- | --- |
| 1 | Double-count in `maybeAutoEvolveStyle` | ✅ Fixed | `latestKey` param threaded from `recordFeedback`, loop skips it, single explicit `+1` — exact count regardless of backend freshness. New regression test covers it. |
| 2 | `context.Background()` in `shouldRequireApproval` | ✅ Fixed | Signature now `(ctx, plan)`, `GetBoolPref` uses request ctx. Test callsites updated. |
| 3 | Inverse trigger for `PrefStyleTerse` | ✅ Fixed | `verbosePhrases` list ("be verbose", "more detail", "longer responses", "more thorough") returns `PrefStyleTerse=false`. Loop is now symmetric. |
| 4 | LLM-judge `MaxTokens=8` too tight | ✅ Fixed | Bumped to 24 with explanatory comment. |
| 5 | Duplicate `maybeStoreUserPreference` on UNRELATED path | ✅ Fixed | `prefStored` flag added in `ChatFromPlatform`; Phase 6 gated on `!prefStored`. |
| 8 | `insertBeforeLast` helper | ✅ Fixed | Extracted; 10 call sites converted (handleChat, handleAnswer, handleSelfSkill, handleRemember, handleToolTask, handleDiagnose, capability/memory/style/diagnose injections). Always allocates fresh backing array. |

Skipped #6 (single-letter regex trip) and #7 (`buildStyleDirective` caching) accepted — both are deferred-until-data calls, reasonable.

Nothing else to flag. LGTM — ready to merge from my side.

PR #15, opened 2026-04-30
Add YouTube video summarization and transcript extraction

srvsngh99:
## Summary
- Add `minikrill youtube <url>` command to extract raw transcripts from YouTube videos
- Add `minikrill youtube summarize <url>` command for AI-generated summaries with key takeaways (overview, bullet points, notable quotes)
- Auto-detect YouTube URLs in the existing `minikrill summarize <url>` command so it routes through the transcript extractor instead of HTML scraping
- Support all YouTube URL formats: standard watch, youtu.be short links, embeds, /v/, and Shorts
- No API key required — extracts captions from the video page's embedded player response
- Prefers English captions, falls back to any available language, and gracefully degrades to video metadata when no captions exist
- 16 unit tests covering video ID extraction, caption URL parsing, title extraction, and JSON unescaping

## New commands
| Command | Description |
|---|---|
| `minikrill youtube <url>` | Extract and print the raw transcript |
| `minikrill youtube summarize <url>` | AI summary with key takeaways |
| `minikrill summarize <youtube-url>` | Also works (auto-detected) |

## Test plan
- [x] `go build ./...` passes
- [x] `go test ./...` passes (all 16 new tests green, no regressions)
- [ ] Manual test: `minikrill youtube <real-url>` prints transcript
- [ ] Manual test: `minikrill youtube summarize <real-url>` returns structured summary
- [ ] Manual test: `minikrill summarize <youtube-url>` auto-detects and uses transcript
- [ ] Test with video that has no captions (should fall back to metadata)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #13, opened 2026-04-30
YouTube video TLDR / summary support


srvsngh99:
## Summary

Add the ability to summarize YouTube videos by extracting captions/transcripts and running them through the LLM. No API key required — YouTube exposes caption tracks in the embedded player config.

## Approach

1. **Detect YouTube URLs** — `youtube.com/watch?v=`, `youtu.be/`, `youtube.com/shorts/`
2. **Extract caption track URL** from the video page's embedded player config JSON
3. **Fetch and parse captions** (XML → plain text)
4. **Summarize via LLM** — reuse existing `Summarize()` pipeline from `internal/content`
5. **Auto-detect in chat** — add "youtube", "video", "summarize this video" triggers to `shouldSearch()` / intent classification

## Where it should work

- **Chat / TUI**: "summarize this youtube video [url]"
- **CLI**: `minikrill web summarize [youtube-url]`
- **Telegram / Discord**: paste a YouTube link and ask for summary

## Files to change

- `internal/content/content.go` — YouTube URL detection, transcript extraction, new `YouTubeSummary()` function
- `internal/agent/agent.go` — add video-related triggers to `shouldSearch()` / `classifyIntent()`
- `internal/plugin/builtins.go` — optional: `youtube` skill
- `internal/chat/telegram.go` — detect YouTube links in messages

## Notes

- No external API key or dependency needed
- Should handle errors gracefully: private videos, no captions available, age-restricted content
- Respect existing content size limits and SSRF protections

PR #30, opened 2026-05-14
Fix silent plan failures, add provenance tags, quick-win hardening

srvsngh99:
First of three PRs in the v0.1.1 cycle. Trust + quick wins. Version stays at 0.1.0.

## Summary

- **Plan execution honesty (#2 / #3 / #4)** — `executeStepWithTools` now sees the original user task on every step. Step 1 used to have only "Previous context" which was empty, so the LLM hallucinated "no link was provided" then the executor logged it `done`. Plus new `NEED_INPUT:` / `BLOCKED:` step statuses (with prose-heuristic fallback) so an LLM that admits it cannot proceed is recorded as such, not as success. Tally line now reports blocked steps separately.
- **Provenance tags (D5)** — new `internal/agent/provenance.go` + tests. Chat and tool-task LLM calls inject a `ProvenanceInstruction` system message asking the model to tag external claims `[web:url] / [mem:key] / [trained] / [guess]`. After generation, `EnforceProvenance` strips any `[web:...]` tag whose URL was not actually fetched this turn and prepends a one-line warning. Per-turn fetch ledger lives on the agent and resets in `ChatFromPlatform`.
- **Search-tool error surfacing (#6)** — `webSearchSkill.Execute` returns real Go errors instead of swallowing them into a string with nil error. `handleToolTask` no longer falls through to plain chat on tool failure (which let the LLM fabricate a "WebSearch needs your green light" excuse) — it shows the actual error.
- **D7 quick wins** — `MallocStackLogging=0` env var on every spawned subprocess + a `noiseFilter` writer wrapping `dive.log` (kills ~3.6k noise lines per long session); codex `--full-auto` → `--sandbox workspace-write` (#8); `SetChannel` no-op deleted (#13); `recordFeedback` correction-trigger tightened so "Nope just find me X" counts as redirection, not dissatisfaction (#14); freshness-keyword router match now requires a co-occurring external-info noun so "what's the current state of my plan" no longer trips a web search (#15).

## Test plan

- [x] `go test ./...` green
- [x] `go vet ./...` clean
- [x] New `EnforceProvenance` tests cover honest tag, fabricated tag, mixed, and no-tags-passthrough
- [ ] Manual: re-feed `summarize this <youtube>` and confirm Step 1 no longer says "no link provided"
- [ ] Manual: trigger a search failure (DDG 429 or no network) and confirm the response shows the real error rather than a hallucinated permission UI
- [ ] Manual: tail `dive.log` after a 30-min session and confirm no `MallocStackLogging` lines appear

### DIFF
--- a/cmd/minikrill/commands.go
+++ b/cmd/minikrill/commands.go
@@ -414,8 +414,14 @@ func diveDaemon() error {
 	}
 
 	child := exec.Command(exe, childArgs...)
-	child.Stdout = out
-	child.Stderr = out
+	// Suppress macOS MallocStackLogging dyld warnings at the source so they
+	// never reach dive.log. Belt-and-suspenders alongside the per-subprocess
+	// env in internal/llm/cli_provider.go.
+	child.Env = append(os.Environ(), "MallocStackLogging=0")
+	// Filter writer drops residual dyld noise lines if any leak through anyway.
+	filtered := newNoiseFilter(out)
+	child.Stdout = filtered
+	child.Stderr = filtered
 	detachCommand(child)
 
 	if err := child.Start(); err != nil {
--- a/cmd/minikrill/log_filter.go
+++ b/cmd/minikrill/log_filter.go
@@ -48,6 +48,24 @@ func (f *noiseFilter) Write(p []byte) (int, error) {
 	return original, nil
 }
 
+// Flush emits any bytes still buffered after the last newline. The daemon
+// process tends to end every line with `\n`, so this is mostly a contract
+// nicety — but if a child exits mid-line (panic, segfault) without a
+// trailing newline, those bytes would otherwise vanish silently.
+func (f *noiseFilter) Flush() error {
+	if f.buf.Len() == 0 {
+		return nil
+	}
+	leftover := f.buf.Bytes()
+	if isNoise(leftover) {
+		f.buf.Reset()
+		return nil
+	}
+	_, err := f.w.Write(leftover)
+	f.buf.Reset()
+	return err
+}
+
 func isNoise(line []byte) bool {
 	for _, pat := range dyldNoisePatterns {
 		if pat.Match(line) {
--- a/cmd/minikrill/log_filter.go
+++ b/cmd/minikrill/log_filter.go
@@ -1,0 +1,58 @@
+package main
+
+import (
+	"bytes"
+	"io"
+	"regexp"
+)
+
+// noiseFilter wraps an io.Writer and drops well-known macOS dyld diagnostic
+// lines that have no signal value. Without it, dive.log fills up with
+// "minikrill(NNNN) MallocStackLogging: can't turn off malloc stack logging
+// because it was not enabled." for every subprocess spawn.
+//
+// The filter operates line-by-line. Bytes received without a trailing newline
+// are buffered until the next write so we never split a line mid-pattern.
+type noiseFilter struct {
+	w   io.Writer
+	buf bytes.Buffer
+}
+
+func newNoiseFilter(w io.Writer) *noiseFilter {
+	return &noiseFilter{w: w}
+}
+
+// dyldNoisePatterns lists regexes that match log lines we always want to drop.
+// Add sparingly — anything matched here is silently discarded.
+var dyldNoisePatterns = []*regexp.Regexp{
+	regexp.MustCompile(`MallocStackLogging:`),
+	regexp.MustCompile(`^objc\[\d+\]: Class .* implemented in both`),
+}
+
+func (f *noiseFilter) Write(p []byte) (int, error) {
+	original := len(p)
+	f.buf.Write(p)
+
+	for {
+		idx := bytes.IndexByte(f.buf.Bytes(), '\n')
+		if idx < 0 {
+			break
+		}
+		line := f.buf.Next(idx + 1) // includes '\n'
+		if !isNoise(line) {
+			if _, err := f.w.Write(line); err != nil {
+				return original, err
+			}
+		}
+	}
+	return original, nil
+}
+
+func isNoise(line []byte) bool {
+	for _, pat := range dyldNoisePatterns {
+		if pat.Match(line) {
+			return true
+		}
+	}
+	return false
+}
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -57,6 +57,7 @@ type KrillAgent struct {
 	subMgr      *SubKrillManager
 	taskStore   *TaskStore
 	taskRunner  *TaskRunner
+	turnFetches *turnFetchLog // per-turn provenance ledger; reset on each ChatFromPlatform call
 	mu          sync.Mutex
 }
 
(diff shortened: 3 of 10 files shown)


srvsngh99:
## Review — PR #30 (trust + quick wins, round 1)

Read the full diff, ran `go build`, `go vet`, `go test ./...` on the head commit (`2fac6f2`). All green. The core moves here are real fixes for real bugs — the planner threading `plan.Task` into every step context, the provenance ledger, the tool-error surfacing, and the freshness-router tightening all target documented gaslighting / silent-success incidents from prior sessions, and the tests in `provenance_test.go` cover the honest/fabricated/mixed/passthrough quadrants.

A few things to address before merge:

### 1. `looksLikeCorrection` looks at the wrong assistant turn (bug)

`internal/agent/agent.go:1175-1183` walks `a.history` backward for the most recent `assistant` message to gate the correction signal. But `recordFeedback` is launched *after* `appendMessage(role=assistant, current response)` in every handler — so the latest assistant message in history is the response we *just* generated to the "no/nope/nah" input, not the prior turn the user was reacting to. The >30-char / no-greeting check is then evaluating the freshly-generated reply, which renders the "previous assistant must be substantive" guard a no-op.

Fix options:
- Pass `prevAssistant` explicitly into `recordFeedback` (capture before `appendMessage`), or
- Skip the last assistant entry: start the scan at `len(a.history) - 2` or skip the trailing assistant before searching backward.

### 2. `detectStepBlocker` heuristics will false-positive on legitimate output

`internal/agent/planner.go:253-265` matches substrings like `"please send the"` and `"still no"` anywhere in the step output. A step that legitimately produces prose such as `"Draft email: please send the user a confirmation when…"` or summarises `"the endpoint returns 'still no results' on empty"` will get marked `need_input` and tank the success tally.

The file's own comment acknowledges this bias is intentional ("better to over-detect blockers than mark hallucinated success as a real outcome"), and I agree with the direction. But I'd at least:
- Anchor the prose heuristics to start-of-output or start-of-line (`strings.HasPrefix(lower, p)` or `^…` regex), since the bug they're catching is the *whole* step output being a refusal.
- Drop `"still no"` — too generic to be safe as a substring match. `NEED_INPUT:` / `BLOCKED:` from the planner-prompt addendum is the durable signal; the prose net should be tight.

### 3. `FormatPlan` doesn't render output for the new `blocked` / `need_input` statuses

`internal/agent/planner.go:92` gates output rendering on `step.Status == "done" || step.Status == "failed"`. The new statuses carry the most user-relevant `step.Output` (what's missing / why it's blocked), but the formatter silently swallows it. Add `"blocked"` and `"need_input"` to that condition so the user actually sees the "needs the URL" message rather than just a `[?]` icon.

### 4. Nit: `contains` / `stringIndex` test helpers in `provenance_test.go`

`internal/agent/provenance_test.go:65-77` reimplements substring search. `strings.Contains` is fine here — no reason to roll our own. Doesn't block anything, just dead weight.

### 5. Nit: `noiseFilter.Write` never flushes a trailing partial line

`cmd/minikrill/log_filter.go` buffers everything after the last `\n` and emits nothing if the child exits without a final newline. For dyld diagnostics that always end in newlines this is fine in practice, but a one-line `Flush()` method (or a `defer` writing leftover bytes when the daemon shuts down) would make the contract clean.

### Things I'm explicitly happy with

- Threading `plan.Task` into per-step context (`planner.go:146-160`) directly cures the "step 1 had no input → LLM hallucinates 'no link provided' → executor logs done" pattern. This is the headline fix and it's correct.
- `ProvenanceInstruction` + post-hoc `EnforceProvenance` is a nice asymmetric guardrail (cheap to add, hard for the model to game). `normalizeURL` symmetry between `Record` and tag-extraction is right.
- `webSearchSkill.Execute` returning real Go errors + `handleToolTask` no longer falling through to plain chat on tool failure (`agent.go:853-862`) is exactly the right shape — kills the "WebSearch needs your green light" fabrication path at the source.
- Removing `SetChannel` and its test: good, it was an unused no-op.

Re-requesting changes for items 1–3; 4 and 5 are nits and won't block.

srvsngh99:
Thanks — addressed all five in `59cd427`. Summary:

**1. `looksLikeCorrection` was scanning the wrong assistant turn** — fixed. `ChatFromPlatform` now snapshots `prevAssistant` *before* the handler appends its response, then passes it explicitly into `recordFeedback`. The function is no longer a method (no longer reads `a.history`) which makes the dependency obvious. New helper `lastAssistantBefore(history, beforeIdx)` does the lookup.

**2. `detectStepBlocker` heuristics were too loose** — fixed. The prose patterns now match against the first 5 lines of output anchored to start-of-line (`strings.HasPrefix` on each trimmed line). A whole-step refusal opens with the refusal phrase; mid-prose quotes like *"Draft email: please send the user a confirmation"* no longer trip. Dropped `"still no"`. `NEED_INPUT:` / `BLOCKED:` tokens remain the durable contract.

**3. `FormatPlan` swallowed output for new statuses** — fixed. The render gate now includes `"blocked"` and `"need_input"` so the user actually sees *what's* missing rather than just `[?]`.

**4. `contains` / `stringIndex` reimplementation** — fixed. Use `strings.Contains` directly.

**5. `noiseFilter.Flush`** — added. Emits leftover bytes when no trailing newline arrives (e.g. child panic / segfault). Drops leftover-noise lines too.

`go build`, `go vet`, `go test ./...` all green on `59cd427`. PTAL.

srvsngh99:
## Approval — PR #30, round 2 ✅

Pulled `59cd427`, re-ran `go build ./... && go vet ./... && go test ./...` — all green. All five items from round 1 are addressed cleanly:

1. **`looksLikeCorrection` now reads the prior turn correctly.** `prevAssistant` is snapshotted at `agent.go:234` (after the user message is appended, before any handler runs) and threaded into `recordFeedback` → `looksLikeCorrection`. The new `lastAssistantBefore` helper at `agent.go:1218` walks backward from the user-message slot, so it lands on the genuine prior assistant turn. Edge case of a brand-new conversation returns `""` → the `<30 char` guard correctly bails. The function dropped its receiver and is now a pure package function, which I like — easier to test, no hidden state.

2. **`detectStepBlocker` prose heuristics anchored.** `planner.go:271-282` now `HasPrefix`-matches on the trimmed first ≤5 lines instead of substring-scanning the whole output. `"still no"` is gone, `"please send the"` is replaced by the three specific variants (`"please send the video"` / `"please send the url"` / `"please send the link"`). The `NEED_INPUT:` / `BLOCKED:` substring check stays as the durable contract. This is exactly the shape I wanted — refusals open the output, legitimate prose that quotes a refusal mid-paragraph no longer trips it.

3. **`FormatPlan` renders blocked/need_input output.** `planner.go:91-96` extended. The whole point of those statuses was to surface what's missing; now it actually does.

4. **Test helper cleanup.** `provenance_test.go` switched to `strings.Contains`. Trivial, done.

5. **`noiseFilter.Flush()` added.** `cmd/minikrill/log_filter.go:51-68` flushes leftover buffered bytes if a child exits mid-line, and still honours the `isNoise` filter. Good shape — note that nobody calls `Flush()` in `commands.go` yet, so the practical effect kicks in only when a future caller invokes it (e.g., during daemon shutdown). Not blocking for this PR, just flagging for follow-up if you want it actually wired up.

### Verdict

LGTM — approving. Headline fixes (planner threading `plan.Task`, provenance ledger, real tool errors, correction-vs-redirection classifier) are sound, tests cover the right quadrants, and the round-1 follow-ups landed without scope creep. Ready to merge as the first of the three v0.1.1 PRs.

srvsngh99:
Thanks for the approval! Filed the `noiseFilter.Flush()` wire-up as a follow-up — agreed it should land before the daemon-shutdown story is real, but happy to keep it scoped out of this PR per your suggestion. Ready to merge whenever.

PR #32, opened 2026-05-14
Personality picker, runtime rename, eyes on self; v0.1.1

srvsngh99:
Third and final PR in the v0.1.1 cycle. Stacks on #31 → #30. Final commit bumps `VERSION` to **0.1.1** and updates `CHANGELOG.md`.

> Stacked on #31 (\`pr2-autonomy-core\`). Once #30 and #31 land, the base will retarget to \`main\`.

### Identity & personality (E1)
- `minikrill init` adds an identity step: pick personality (Buddy / Jarvis / Friday / Samantha / Tars / Krill) and name the agent. Default for new users is Buddy.
- `internal/brain/soul.go`: `genericSystemPrompt` is the new neutral fallback. The leaky `"%s - a custom Mini Krill personality"` line replaced with `"%s - your AI assistant"`. Krill voice becomes one personality, not the unconditional default.
- `/name Edwin` slash command + natural-language triggers (\"call you Edwin\", \"your name is Sage\", \"from now on you're …\") on every interface. Allowlist sanitiser blocks injection-shaped names.
- `/persona <description>` overlay: layered system-prompt directives in `~/.mini-krill/personalities/_overlay.yaml`. Manage with `/persona list | undo | reset`. Lightweight injection check refuses overlay attempts that look like prompt overrides.
- `/personality <name>` switches voice mid-session.

### Emoji control (E2)
- `cfg.Brain.EmojiStyle` ∈ `{none, sparse, playful}` (default `sparse`; `playful` for existing krill users to preserve voice). Output filter strips or trims emoji.
- `/emoji none|sparse|playful` + natural-language triggers (\"stop with the emojis\", \"more emojis\").
- Approval/rejection regex extended with emoji acks (👍 ✅ 👎 ❌ 🛑 etc.). Telegram thumbs-up now approves plans without an LLM round-trip.
- `recordFeedback` reads emoji sentiment so the affinity loop responds to how Telegram users actually communicate.

### Memory & scrollback (E3 + Issue #19)
- `ConversationStore.Search(channel, query, since, limit)` — scans the entire JSONL, not just the sliding window.
- `/recall last N` / `/recall yesterday` / `/recall today` / `/recall about <topic>` + natural-language triggers (\"what did I say yesterday\", \"go back to our last 5 messages\", \"remember when we discussed …\").
- The Telegram `ReplyToMessage` capture from the audit (E3.1) is **deferred** to a follow-up to keep this PR scoped — covered by the existing audit and ready for a small dedicated change.

### Eyes on self (E4)
- `self:read-code` — `read` / `grep` / `tree` over the agent's own source. Sandboxed to the repo root (resolved via `MINIKRILL_REPO` env var or go.mod walk). Refuses paths matching the secret allowlist (`config.yaml`, `.env`, `*.key`, etc).
- `self:read-logs` — `tail` / `grep` / `errors` over `~/.mini-krill/logs/`.
- Both strictly read-only. Self-modification path (`self:propose-patch`) is intentionally out of scope for this PR.

### Version bump
- `VERSION`: 0.1.0 → **0.1.1**
- `CHANGELOG.md`: new `[0.1.1] - 2026-05-15` section covering PRs #30, #31, and this one.

## Test plan

- [x] `go test ./...` green
- [x] `go vet ./...` clean
- [ ] `minikrill init` on a fresh data dir: confirm wizard asks personality + name + emoji style and the resulting `config.yaml` has `agent_name`, `personality`, and `emoji_style` set.
- [ ] On Telegram: send \"call you Sage\" — confirm the next reply uses Sage and `agent_name: Sage` lands in `config.yaml`.
- [ ] On TUI: type `/persona drop the metaphors` then chat — confirm metaphor density drops in subsequent replies.
- [ ] `/emoji none` then send a message that would normally have emoji — confirm strip.
- [ ] After 10+ turns: type \"what did I say about <topic>\" — confirm the agent quotes a real prior turn.
- [ ] Type \"check your logs for errors today\" — confirm structured log output, no fabrication.
- [ ] Type \"show me your code for the freshness matcher\" — confirm `self:read-code` returns the actual `router.go` lines.

## Deferred (follow-ups, not in this PR)
- E3.1 — Telegram/Discord `ReplyToMessage` capture + `platform_msgid.jsonl` mapping
- D6 — episodic session-summary consolidation on idle
- E4.3 — `self:explain-decision` (composes affinity store + read-code + read-logs)


srvsngh99:
**Rebased onto current \`pr2-autonomy-core\` head (\`43138a3\`).** Force-pushed to \`29b9e56\`. Now exactly 1 commit on top of PR #33's approved state — diff cleanly shows just the identity / memory / eyes-on-self work.

What changed during the rebase:
- Dropped \`lastCluster\` from \`KrillAgent\` (PR #33 round-2 review removed it; pre-rebase #32 still had it). The struct merge took the post-fixup shape (no \`lastCluster\`, plus this PR's \`cachedEmojiStyle\`).
- No code changes needed beyond the struct conflict resolution — everything else applied cleanly.

Verified:
- \`go build ./...\` green
- \`go vet ./...\` clean
- \`go test ./...\` all packages pass
- 1-commit delta against \`pr2-autonomy-core\` (\`git log origin/pr2-autonomy-core..HEAD\` shows just \`29b9e56\`)

**Stack note:** This PR currently bases on \`pr2-autonomy-core\`. Once PR #33 squash-merges into \`main\`, GitHub will likely auto-close this one too (same pitfall #31 hit). I'll watch for that and recreate against \`main\` immediately so review can continue without interruption — the diff against \`main\` will be identical content since #33's changes will be in main by then.

Ready for review.

srvsngh99:
## Review — PR #32 (identity, recall, eyes-on-self; v0.1.1 bump, round 1)

Read the full diff (1648 lines, 13 files), retargeted base to `main` (the original `pr2-autonomy-core` base was the now-stale stacked branch), ran `go build ./... && go vet ./... && go test ./...` on the head (`29b9e56`) — all green. Big PR; the identity/persona/recall/self-eyes surface is well-scoped and the path sandboxing in `safeRepoPath` is correct.

A few substantive items to address:

### 1. Rename regex greedy-captures past the name (identity.go:13-23)

```go
regexp.MustCompile(`(?i)\b(?:i'?ll )?call you ([A-Za-z0-9 '\-]{1,40})\b`),
```

`[A-Za-z0-9 '\-]{1,40}` is greedy and includes spaces, so "call you Mr Smith from now on" captures the entire tail — `AgentName` becomes `"Mr Smith from now on"`. Same problem for the other triggers ("your name is Sage but only on Tuesdays" → `"Sage but only on Tuesdays"`).

Fix shape: require the capture to end at a comma/period/end-of-string or a closed set of stop-words, e.g.

```go
regexp.MustCompile(`(?i)\bcall you ([A-Za-z][A-Za-z0-9'\-]*(?:\s+[A-Za-z][A-Za-z0-9'\-]*)?)(?:[\s,.!?]|$)`)
```

— or capture one or two name-shaped tokens explicitly. The current rule means any prose-tail after the name silently becomes part of the agent's display name on disk.

### 2. `trimAfter` is a no-op so `/recall yesterday` includes today's turns (recall.go:135-141)

```go
func trimAfter(msgs []core.Message, _ time.Time) []core.Message {
    return msgs
}
```

The comment acknowledges this — `core.Message` has no timestamp, so the upper bound can't be applied client-side. But the semantics of "yesterday" require excluding today. As shipped, `/recall yesterday` returns yesterday-or-later, which silently includes today's turns and is the opposite of what the user asked for.

Two clean fixes:
- Extend `ConversationStore.Search` to accept `until time.Time` and filter server-side (`conversations.go:166` already iterates `turn.Timestamp`).
- Or have Search return turns paired with timestamps and let recall.go filter.

Either way, the `trimAfter` stub should not survive as-is — it makes "yesterday" lie.

### 3. Persona overlay is re-read from disk on every chat turn (agent.go:940-947)

`buildOverlayDirective` calls `OverlayDirectives()` which `os.ReadFile`s `~/.mini-krill/personalities/_overlay.yaml` every time `handleChat` runs. Cheap per call but pointless I/O on a hot path. Cache the parsed overlay in the agent struct and invalidate from `addPersonaDirective`/`saveOverlay` (and `/persona reset/undo`). Same pattern as `cachedEmojiStyle` already uses.

### 4. `detectEmojiPreference` substring-matches in unintentional contexts (identity.go:359-373)

`strings.Contains(lower, "no emoji")` fires on "I'm writing a paragraph with no emoji characters" — and silently flips the user's persistent style to `none`. Same shape for the other branches. This setter mutates `~/.mini-krill/config.yaml`, so a single false-positive sticks across sessions.

Tightening options:
- Require an imperative frame: `(?i)\b(?:use|prefer|want|stop|no more|drop|please)\s+(?:no |fewer |more |sparse |playful )?emoji`.
- Or only match when the input is short (≤ 8 words) — same heuristic the correction-vs-redirection classifier uses.

A persistent config mutation hidden behind a substring match needs a tighter trigger.

### 5. `RenameAgent` truncates before validating (identity.go:50-66)

```go
if len(name) > 40 {
    name = name[:40]
}
if !safeNameRE.MatchString(name) {
    return "names can only contain ..."
}
```

If the user submits 60 chars where the first 40 are allowlist-clean and the last 20 contain junk, truncation strips the junk and the allowlist passes silently. Not a security hole (every allowed char is safe), but a UX cliff — the user typed garbage and got "Got it, I'll go by ..." back with the garbage silently dropped.

Fix: validate the *full* trimmed input first, return the "can only contain ..." error if it fails, and only then enforce the 40-char display cap.

### 6. No tests for any of the new code (especially the sandboxing) — 1341 lines, 0 test files

`git diff main..HEAD --name-only | grep test` returns nothing. The pieces I'd most want unit-tested before merge are the security-shaped ones:

- `safeRepoPath` — escape attempts like `../../etc/passwd`, absolute-path inputs, MINIKRILL_REPO unset path.
- `safeLogsPath` — same.
- `addPersonaDirective` — each banned-phrase trigger.
- `applyEmojiStyle` — none/sparse/playful round-trip with and without emoji clusters.
- `detectRenameRequest` — including the greedy-capture cases from item #1.
- `ConversationStore.Search` — channel filter, since filter, query filter, the limit-as-trailing-window behaviour.

`detectRecallRequest` and `detectEmojiPreference` are also pure functions begging for table-driven tests.

Skipping tests on the affinity/cluster PR (#33) would have hidden the narration bug we caught in review there; the same risk applies to a much bigger surface here. The sandboxing in particular is exactly the kind of thing that needs a test the day it's written — a regression on `safeRepoPath` exposes the user's source tree silently.

### Nits

- **`secretPathPatterns` uses unbounded substring match** (`self_eyes.go:30-33`). A file innocently named `notconfig.yaml.txt` gets refused. Erring on the safe side is fine; just calling out.
- **CHANGELOG framing vs. config default mismatch.** The blurb says "Krill voice becomes one personality, not the unconditional default" — true for new installs (`buddy` default in the init wizard) — but `fillDefaults` only sets `EmojiStyle = "playful"` when `Personality == "krill"`. Existing users who land in `fillDefaults` without setting `cfg.Agent.Personality` will get `sparse`, which is probably correct, but the section reads as if the krill default is preserved across the board. Minor.
- **`core/types.go` whitespace-only churn.** Removing a space in `// system, user, assistant` and re-aligning the option setters adds 5 cosmetic lines to the diff. Not blocking — just noisy.

### Things I explicitly like

- **`safeRepoPath` sandbox is correct.** Clean → Abs → `HasPrefix(rootAbs + sep)` is the right pattern; the in-loop re-validation inside `repoGrep` after `filepath.Glob` is the load-bearing part. Belt-and-suspenders with `secretPathPatterns` on top is good defence-in-depth.
- **Persona overlay injection allowlist** (`addPersonaDirective`'s banned phrases) is exactly the lightweight check a "user can rewrite my style" feature needs. Refusing "ignore previous" / "system:" / "api key" / "dump memory" closes the easy attack surface.
- **Conversation `Search` scanner buffer at 4MB max** survives reasonably long turns without choking.
- **Emoji acks in approve/reject regex** is a nice Telegram-native ergonomic win — and `recordFeedback` reading 👍/👎 means the affinity loop responds to how mobile users actually communicate. Tasteful.
- **`/personality` listing filters `_`-prefixed files** so `_overlay.yaml` stays out of the user-facing list. Good defensive UX.
- **CHANGELOG section actually documents what changed** instead of just listing files. v0.1.1 bump is in the right PR (final of the three-PR stack).

### Verdict

Re-requesting changes for items 1–5 and the missing-tests gap (#6). Items 7–9 are nits and won't block.

srvsngh99:
**Retargeted to `main` and rebased.** PR #33 squash-merged at 20:21, so I:

1. Rebased `pr3-identity-memory-self-bump` onto current `origin/main` (which now includes #33's autonomy work via squash `1a054ca`).
2. Skipped 3 commits whose contents were already upstream in the squash (`0277c55`, `67d8173`, `43138a3`).
3. Force-pushed to `025a577`.
4. Switched PR base from `pr2-autonomy-core` → `main`.

Now exactly **1 commit on top of main**, 13 files changed, identity / personality / emoji / recall / eyes-on-self isolated. Mergeable, CI re-running.

`go build`, `go vet`, `go test ./...` all green on `025a577`.

Ready for review.

srvsngh99:
Thanks — addressed all 5 blockers + the test gap in \`7d41451\`.

**1. Rename regex stops at name boundary.** Capture is now \"one or two CAPITAL-letter-led name tokens\" with an explicit stop-char/EOL boundary. Lowercase common words terminate the match. \`(?i:...)\` scopes the case-insensitive flag to the trigger phrase only, so the capital-letter requirement holds. \`call you Mr Smith from now on\` → \`Mr Smith\`. \`your name is Sage but only on Tuesdays\` → \`Sage\`. New \`TestDetectRenameRequest_StopsAtBoundary\` pins both cases.

**2. \`/recall yesterday\` actually means yesterday.** \`ConversationStore.Search\` gained an \`until time.Time\` upper-bound parameter, filtered server-side. Killed the dead client-side \`trimAfter\` stub. New \`TestSearch_RespectsSinceAndUntilWindow\` seeds two-days-ago / yesterday / today turns and asserts the yesterday window returns only the yesterday entry.

**3. Persona overlay cached on the agent.** \`cachedOverlayDirs\` + \`cachedOverlayInit\` fields, invalidated from \`addPersonaDirective\` and the \`/persona reset\` / \`/persona undo\` handlers. \`buildOverlayDirective\` is now a method that hits the cache. \`TestOverlayCacheInvalidation\` confirms end-to-end: empty cache → add directive → next read sees the new directive.

**4. \`detectEmojiPreference\` tightened.** Now requires an imperative frame (\`use|prefer|want|stop|drop|switch to|please use|...\`) before an emoji-style noun, bounded by ≤14 words on the message. Three plausible long-prose false-positive cases are pinned in \`TestDetectEmojiPreference_NoFalsePositives\` (\"I'm writing a paragraph with no emoji characters in it\" etc).

**5. \`RenameAgent\` validates before truncating.** Reordered: trim → validate full input against the allowlist → only then cap at 40. \`TestRenameAgent_RejectsJunkBeforeTruncating\` pins this with a \"40 clean chars + 20 junk\" input, and \`TestRenameAgent_AcceptsCleanLongInput\` confirms the legitimate truncation path still works.

**6. Tests added (1648 → 2025 lines, +377 of tests):**
- \`identity_test.go\` — rename triggers (basic + boundary regression + no-match), RenameAgent (4 cases inc. injection), emoji preference (matches + false-positive regression), applyEmojiStyle (none/sparse/playful + plain passthrough).
- \`recall_test.go\` — detectRecallRequest for lastN/yesterday/today/topic/no-match. Caught a real bug while writing: \`recallTriggers\` order had topic patterns before yesterday/today, so \"what did I say yesterday\" was matching topic=\"yesterday\". Fixed by moving specific patterns first.
- \`persona_test.go\` — addPersonaDirective injection allowlist (8 banned phrases), legit acceptance + overlay-file write check, overlay cache invalidation end-to-end.
- \`self_eyes_test.go\` — safeRepoPath escape rejection (4 traversal/absolute cases), secret allowlist (5 secret-shaped paths), clean-relative happy path. safeLogsPath escape rejection + bare-filename happy path.
- \`conversations_test.go\` — Search since+until (the yesterday regression), query filter, channel filter, limit-as-trailing-window.

Nits 7-9 left as-is per your guidance (non-blocking; the secretPathPatterns substring match is intentional defence-in-depth).

\`go build\`, \`go vet\`, \`go test ./...\` all green on \`7d41451\`. PTAL.

srvsngh99:
## Approval — PR #32, round 2 ✅

Pulled `7d41451` (force-push replacing `025a577`), re-ran `go build ./... && go vet ./... && go test ./...` — all green. Also ran the round-1-fix-targeted tests explicitly (TestDetectRename*, TestRenameAgent*, TestDetectEmojiPreference*, TestApplyEmojiStyle*, TestDetectRecallRequest*, TestPersona*, TestSafeRepoPath*, TestSafeLogsPath*, TestSearch*) — every one passes. Each round-1 item is addressed cleanly:

### 1. Rename regex stops at the name boundary — fixed

`identity.go:27-32` introduces `nameTokenRE = [A-Z][A-Za-z0-9'\-]*` requiring a capital-letter start, composed into `nameCaptureRE = ` one or two tokens, followed by `nameStopRE = (?:[\s,.!?]|$)`. The case-insensitive flag is scoped to the trigger phrase via `(?i:...)` so the capital-letter requirement on the name itself is preserved. "call you Mr Smith from now on" captures `Mr Smith` (lowercase `from` terminates the match); "your name is Sage but only on Tuesdays" captures `Sage`. The boundary regression test in `identity_test.go` pins both.

### 2. `/recall yesterday` actually means yesterday — fixed

`core/types.go:130` — `ConversationStore.Search` signature now `(channel, query, since, until time.Time, limit)`. The filter is applied server-side at `conversations.go:175-177` (`if hasUntil && !turn.Timestamp.Before(until) { continue }`). All `recall.go` callers updated; the dead `trimAfter` stub is gone. `TestSearch_RespectsSinceAndUntilWindow` pins the exact regression: a fixture with two-days-ago, yesterday, and today turns under the yesterday window returns only the yesterday entry.

Bonus catch in the same commit: `recallTriggers` order was reshuffled so `yesterday`/`today` checks fire *before* the catch-all `topic` regex (otherwise "what did I say yesterday" would match the topic regex with `topic="yesterday"`). That's the kind of thing the new `TestDetectRecallRequest_Yesterday` would have caught even if I hadn't flagged item #2.

### 3. Persona overlay cached — fixed

`agent.go:62-65` — new `cachedOverlayDirs` + `cachedOverlayInit` fields. `agent.go:981-998` adds the `overlayDirectives()` reader and `invalidateOverlayCache()` invalidator. The cache is invalidated from `addPersonaDirective` (`identity.go:184`) and both `/persona reset` (`identity.go:208`) and `/persona undo` (`identity.go:217`). End-to-end cache-then-invalidate test in `persona_test.go`.

### 4. `detectEmojiPreference` imperative-only — fixed

`identity.go:438-462` replaces the substring checks with `emojiPrefRegexes` requiring imperative verb prefixes (`use|prefer|want|stop|drop|cut|kill|switch to|please use`). Length bound `if len(strings.Fields(input)) > 14 { return "" }` caps the trigger to short imperative messages. `TestDetectEmojiPreference_NoFalsePositives` pins three plausible long-prose negatives ("I'm writing a paragraph with no emoji characters" and friends) — all return empty, so config.yaml stays untouched.

### 5. `RenameAgent` validates before truncating — fixed

`identity.go:75-92` — trim → empty check → allowlist check on the full input → THEN cap to 40. The old order ("truncate then validate") would silently strip garbage past position 40 and accept the clean prefix; now `<40 clean> <20 junk>` gets the "names can only contain ..." rejection. `TestRenameAgent_RejectsInjection` covers it.

### 6. Test coverage — comprehensive (565 lines added across 5 new files)

- **`identity_test.go`** (194 lines) — detectRenameRequest happy path + the greedy-capture regression + no-match; RenameAgent four-quadrant matrix (validate-before-truncate, accept-clean-long, reject-empty, reject-injection); detectEmojiPreference imperative-positive + long-prose-negative; applyEmojiStyle for none/sparse/playful + no-emoji passthrough.
- **`persona_test.go`** (87 lines) — addPersonaDirective injection allowlist (per-phrase), legit-acceptance + overlay file write verification, empty-input rejection, cache invalidation end-to-end.
- **`recall_test.go`** (61 lines) — detectRecallRequest for lastN/yesterday/today/topic/no-match. Catches the trigger-order issue that the round-2 commit fixed.
- **`conversations_test.go`** (108 lines) — Search with since+until window (the yesterday regression test), query filter, channel filter, limit-as-trailing-window. This is the most load-bearing addition — pins the v0.1.1 `/recall yesterday` semantics in code.
- **`self_eyes_test.go`** (115 lines) — `safeRepoPath` traversal-rejection (including absolute-path attempts), secret allowlist rejection, clean-relative happy path; `safeLogsPath` escape rejection + bare-filename happy path. `TestSafeRepoPath_NoRoot` gracefully skips when the runtime fallback finds `go.mod` (correct test hygiene for env-dependent behavior).

The sandboxing-test set in particular is exactly what was missing in round 1 — a regression on `safeRepoPath` would have silently exposed the user's source tree. Now it can't slip past CI.

### Nits

The CHANGELOG framing / `secretPathPatterns` substring / `core/types.go` whitespace nits from round 1 weren't directly addressed; all three were explicitly marked "won't block" and are correctly left to follow-ups.

### Verdict

LGTM — approving. All 6 items resolved cleanly. The tightening of `detectEmojiPreference` (length cap + imperative frame), the `until`-bounded `Search` API, and the sandbox/escape unit tests are the kind of fixes that pay back interest later — they don't just patch the reported regression, they pin the contract. Ready to merge as the final PR of the v0.1.1 cycle.

Reminder for merge: PR 32 is the tail of the stack now (PR 30 and PR 33 already merged), so `--delete-branch` is safe — no downstream PR will be auto-closed by deleting `pr3-identity-memory-self-bump`.

srvsngh99:
Thanks for the approval! Heads-up: there's one commit on top of \`7d41451\` (which you pulled for review) — \`53d0a8a\` fixes 6 CI lint errors that surfaced after the round-1 push:

- 2× errcheck on \`fmt.Sscanf\` in \`self_eyes.go\` (explicitly discarded with \`_, _ =\`)
- 3× SA1019 on deprecated \`strings.Title\` (replaced with a 6-line \`capitalizeFirstASCII\` helper — not worth pulling \`golang.org/x/text/cases\` for a single-character upcase)
- 2× unused warnings on \`genericSystemPrompt\` / \`genericSoul\` — wired both into \`LoadPersonalityByName\`'s missing-yaml fallback. **Bonus catch:** this closes a real semantic hole that wasn't in the round-1 review — a user who set \`agent.personality: foo\` in config without a matching \`foo.yaml\` file used to fall through to the krill default. Now they get a clean name-templated neutral soul.

\`go build\`, \`go vet\`, \`go test ./...\` all green on \`53d0a8a\`. CI re-running.

Ready to merge whenever — agreed \`--delete-branch\` is safe for this one (PR 30 and PR 33 are already merged, no downstream stack to break).

PR #17, opened 2026-04-30
Fix Plan: prefix leak, compact chat header, Telegram timeout

srvsngh99:
## Summary
- **Strip internal reasoning prefixes** (`Plan:`, `Thinking:`, `Internal:`) from LLM responses so users only see the actual reply — fixes the "Plan: great, then wait for your next command" leak visible in both TUI and Telegram
- **Compact chat header** — Chat tab now uses a single-line branding bar instead of the 14-line ASCII art banner, giving the chat viewport nearly full screen space
- **Telegram reliability** — Added 90s per-message timeout to prevent a slow/hung LLM from blocking all Telegram messages, fixed reaction swap that checked an always-nil error, and surfaced bot startup failures in TUI mode
- **Better message formatting** — Rewrote `wordWrap` to preserve paragraph breaks, list items, indentation, and blank-line separators for structured LLM responses

## Test plan
- [ ] Run `krill chat` and send "hi" — response should NOT start with "Plan:"
- [ ] Run `krill tui` and switch to Chat tab — header should be a single compact line, chat viewport should fill the screen
- [ ] Switch to Dashboard tab — full ASCII art banner should still appear
- [ ] Send a long structured message in chat — paragraphs/lists should render with proper line breaks
- [ ] If Telegram is configured, send a DM — bot should respond without "Plan:" prefix
- [ ] `go test ./...` passes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #11, opened 2026-04-30
Upgrade chat to TUI and improve setup guidance

srvsngh99:
## Summary
- **Chat UX overhaul**: `minikrill chat` now launches the full Bubble Tea TUI directly on the Chat tab — user messages appear right-aligned in bubbles, bot replies left-aligned in green bubbles, with a styled input bar at the bottom. Replaces the bare `you > ` / `krill > ` stdin loop.
- **Fix "command not found"**: Init "Next steps" now shows `go run ./cmd/minikrill ...` commands (the binary isn't in PATH after a fresh clone) and adds a `go install` tip.
- **Telegram/Discord setup guidance**: Shows how to get bot tokens inline (@BotFather for Telegram, Discord developer portal for Discord), explains that `minikrill dive` is required to start the bots, and adds `dive` to "Next steps" when bots are enabled.

## Test plan
- [ ] Run `go run ./cmd/minikrill init` — verify @BotFather hint appears after enabling Telegram, Discord portal hint for Discord, and "run dive to start" messages
- [ ] Verify "Next steps" shows `go run ./cmd/minikrill` commands and `go install` tip
- [ ] Verify `go run ./cmd/minikrill chat` opens the TUI on the Chat tab with bubble-style messages
- [ ] Verify `go run ./cmd/minikrill tui` still opens on the Dashboard tab (default)
- [ ] Run `go test ./...` — all tests pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #33, opened 2026-05-14
Add task-type plan-affinity learning + autonomy_floor config

srvsngh99:
Replaces the auto-closed #31 (its base \`pr1-trust-and-quickwins\` was deleted when #30 merged). Same diff, rebased onto current \`main\`.

Second of the v0.1.1 cycle. Version stays at 0.1.0.

## Summary

- **\`autonomy_floor\` config** replaces \`plan_approval={always,never,auto}\` with \`{observe,suggest,act,evolve}\`. Default flips to \`act\`. Migration at config load + a second migration in \`agent.New()\` for hand-built \`AgentConfig\` (tests, embedders). Old YAML configs continue to work.
- **Task-type clustering (D1)** — \`internal/agent/cluster.go\`. Maps a user request to \`verb/object\` (sha1[:12]). Lossy on purpose: ~30 stable clusters across 3000 unique inputs.
- **Plan-affinity store (D2)** — \`internal/agent/affinity.go\`. Per-cluster \`[0.05, 0.95]\` score persisted in the existing Memory backend under \`affinity:cluster:<id>\`. Bayesian-ish pull-toward-extreme update rule. Below 3 samples yields \`DecisionUseLegacy\` and the old heuristics decide.
- **Outcome detection (D3)** — \`handlePendingPlan\` credits APPROVED / MODIFIED / REJECTED / IGNORED to the pending cluster, asynchronously. New requests of the same shape need less ceremony over time.
- **Bug #5 / #11 fix folded in** — "No need for taking my approval on this kind of things" mid-plan now flips the pref AND treats the current plan as approved AND credits the cluster, all in one shot. No more "Yea" → "Yes" double-tap loop.
- **Issue #30 — narrate every flip** — when a cluster crosses the calibration threshold into auto-run territory, the next response appends a one-line narration. Autonomy is no longer silent.
- **Issue #17 — ask-the-user pattern** — when the planner's first step is "Ask the user for X", \`handleTask\` now skips the plan and asks the question directly.

## Behaviour changes worth noting

- \`PlanApproval=never\` no longer bypasses destructive checks. No autonomy floor authorises destructive actions silently. \`TestShouldRequireApprovalNever\` updated to test the reversible-plan path instead.
- Default \`autonomy_floor=act\` means most plans run without an approval gate. Operators who want the old gate behaviour can set \`autonomy_floor: suggest\` in \`~/.mini-krill/config.yaml\`.

## Test plan

- [x] \`go test ./...\` green
- [x] \`go vet ./...\` clean
- [ ] Manual: send "summarize this <youtube url>" 5 times, approving each plan. Confirm the 6th run auto-executes without a plan and the 5th carries the narration line.
- [ ] Manual: send a destructive plan with \`autonomy_floor=act\`. Confirm it still gates.
- [ ] Manual: with a pending plan, type "no need for approval here" and confirm the plan executes immediately AND the typed pref flips.
- [ ] Manual: ask "summarize this video" without a link. Confirm the agent asks for the link directly instead of generating a 5-step "I will ask the user, then…" plan.

### DIFF
--- a/internal/agent/affinity.go
+++ b/internal/agent/affinity.go
@@ -1,0 +1,216 @@
+// Package agent — affinity.go maintains a per-cluster "should I plan this?"
+// score that the agent updates from outcomes (approved / modified / rejected /
+// ignored / thanked / fixed).
+//
+// The score is stored in the existing core.Memory under reserved keys
+// `affinity:cluster:<id>`. After ~5 samples per cluster the agent can decide
+// autonomously whether to plan-then-act, plan-in-parallel, or just act, with
+// no user intervention.
+package agent
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/srvsngh99/mini-krill/internal/core"
+	log "github.com/srvsngh99/mini-krill/internal/log"
+)
+
+// AffinityKeyPrefix reserves a memory key namespace so listing/searching the
+// regular memory store skips affinity entries.
+const AffinityKeyPrefix = "affinity:cluster:"
+
+// AffinityOutcome is the discrete event types the store understands.
+type AffinityOutcome string
+
+const (
+	OutcomeApproved AffinityOutcome = "approved"
+	OutcomeModified AffinityOutcome = "modified"
+	OutcomeRejected AffinityOutcome = "rejected"
+	OutcomeIgnored  AffinityOutcome = "ignored"
+	OutcomeThanked  AffinityOutcome = "thanked" // user happy after auto-execute → less planning
+	OutcomeFixed    AffinityOutcome = "fixed"   // user corrected after auto-execute → more planning
+)
+
+// PlanDecision is what the affinity store recommends for a fresh request.
+type PlanDecision string
+
+const (
+	DecisionPlanThenAct    PlanDecision = "plan_then_act"    // generate plan, show as FYI, auto-execute
+	DecisionPlanInParallel PlanDecision = "plan_in_parallel" // generate plan, start execution in parallel
+	DecisionJustAct        PlanDecision = "just_act"         // skip plan generation entirely
+	DecisionUseLegacy      PlanDecision = "use_legacy"       // not enough samples, fall back to old gate
+)
+
+// minSamples is the calibration floor — below this the store won't recommend
+// anything stronger than DecisionUseLegacy.
+const minSamples = 3
+
+// ClusterAffinity is the persisted record per cluster.
+type ClusterAffinity struct {
+	ClusterID    string    `json:"cluster_id"`
+	Verb         string    `json:"verb"`
+	ObjectClass  string    `json:"object_class"`
+	PlanScore    float64   `json:"plan_score"` // [0.05, 0.95], starts at 0.5
+	SampleCount  int       `json:"sample_count"`
+	LastUpdate   time.Time `json:"last_update"`
+	LastOutcomes []string  `json:"last_outcomes"` // ring buffer, max 10
+}
+
+// AffinityStore wraps a core.Memory with affinity-typed accessors.
+type AffinityStore struct {
+	mem core.Memory
+}
+
+// NewAffinityStore wires the store to the existing memory backend. The mem
+// store can be nil for tests; in that case all reads return zero values and
+// all writes are no-ops.
+func NewAffinityStore(mem core.Memory) *AffinityStore {
+	return &AffinityStore{mem: mem}
+}
+
+// Get returns the affinity record for cluster id, or a fresh-default if not
+// previously seen. The returned record is safe to mutate; call Put to persist.
+func (s *AffinityStore) Get(ctx context.Context, c TaskCluster) ClusterAffinity {
+	rec := ClusterAffinity{
+		ClusterID:   c.ID,
+		Verb:        c.Verb,
+		ObjectClass: c.ObjectClass,
+		PlanScore:   0.5,
+		LastUpdate:  time.Now(),
+	}
+	if s.mem == nil {
+		return rec
+	}
+	entry, err := s.mem.Recall(ctx, AffinityKeyPrefix+c.ID)
+	if err != nil || entry == nil || entry.Value == "" {
+		return rec
+	}
+	if err := json.Unmarshal([]byte(entry.Value), &rec); err != nil {
+		log.Warn("affinity record corrupt, resetting", "cluster", c.ID, "error", err)
+		return ClusterAffinity{
+			ClusterID:   c.ID,
+			Verb:        c.Verb,
+			ObjectClass: c.ObjectClass,
+			PlanScore:   0.5,
+			LastUpdate:  time.Now(),
+		}
+	}
+	return rec
+}
+
+// Put persists the affinity record.
+func (s *AffinityStore) Put(ctx context.Context, rec ClusterAffinity) error {
+	if s.mem == nil {
+		return nil
+	}
+	rec.LastUpdate = time.Now()
+	data, err := json.Marshal(rec)
+	if err != nil {
+		return err
+	}
+	return s.mem.Store(ctx, core.MemoryEntry{
+		Key:        AffinityKeyPrefix + rec.ClusterID,
+		Value:      string(data),
+		Tags:       []string{"affinity", "system"},
+		Scope:      "system",
+		Source:     "affinity-store",
+		CreatedAt:  rec.LastUpdate,
+		AccessedAt: rec.LastUpdate,
+	})
+}
+
+// Update applies an outcome to a cluster's score using a simple
+// pull-toward-extreme rule, persists, and returns the post-update record.
+//
+//	APPROVED  → score += 0.10 * (1 - score)
+//	MODIFIED  → score += 0.05 * (1 - score)
+//	REJECTED  → score -= 0.20 * score
+//	IGNORED   → score -= 0.15 * score
+//	THANKED   → score -= 0.10 * score      (user happy without plan = plan was unneeded)
+//	FIXED     → score += 0.15 * (1 - score) (user corrected after auto-act = plan would help)
+//
+// Result is clamped to [0.05, 0.95] so a cluster can always recover.
+func (s *AffinityStore) Update(ctx context.Context, c TaskCluster, outcome AffinityOutcome) (ClusterAffinity, error) {
+	rec := s.Get(ctx, c)
+	score := rec.PlanScore
+	switch outcome {
+	case OutcomeApproved:
+		score += 0.10 * (1 - score)
+	case OutcomeModified:
+		score += 0.05 * (1 - score)
+	case OutcomeRejected:
+		score -= 0.20 * score
+	case OutcomeIgnored:
+		score -= 0.15 * score
+	case OutcomeThanked:
+		score -= 0.10 * score
+	case OutcomeFixed:
+		score += 0.15 * (1 - score)
+	default:
+		return rec, fmt.Errorf("unknown affinity outcome: %s", outcome)
+	}
+	if score < 0.05 {
+		score = 0.05
+	} else if score > 0.95 {
+		score = 0.95
+	}
+	rec.PlanScore = score
+	rec.SampleCount++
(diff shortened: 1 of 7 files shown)


srvsngh99:
## Review — PR #33 (autonomy_floor + task-type affinity, round 1)

Read the full diff (1093 lines, 7 files), checked out the head (`0277c55` — the rebased single-commit) and ran `go build ./... && go vet ./... && go test ./...` clean. This is the replacement for the auto-closed #31; same content rebased onto current `main`.

Solid PR overall — clean separation between `cluster.go` (derivation), `affinity.go` (persistence), and `agent.go` (wiring), and the destructive-plans-always-gate safety change is the right call. A few things to address before merge:

### 1. `maybeAppendAffinityNarration` only fires at the exact moment `SampleCount == minSamples` (agent.go:541-549)

```go
if rec.SampleCount == minSamples && rec.PlanScore > 0.7 {
    return response + fmt.Sprintf(...)
}
```

This narrates only when the *exact* transition into auto-run territory happens at the 3rd sample. But a cluster that takes 5 approvals to cross 0.7 (because earlier outcomes were MODIFIED, or there was a stray REJECTED) will silently slip into auto-run with no narration. The Issue #30 motivation ("autonomy without narrated learning is a black box") wants the narration to track *crossing the 0.7 threshold*, not a one-shot at sample 3.

Fix shape: persist `LastNarratedScore` (or `WasAutoRun`) on the record, fire narration once when the boolean flips from false → true. Or pass the pre-update score into the narration check so you can detect a true crossing.

### 2. `PlanDecision` tiers collapse to one behavior (agent.go:710-718)

`shouldRequireApproval`'s affinity branch returns `false` for `PlanThenAct`, `PlanInParallel`, *and* `JustAct` — they all just disable the gate. But the type docstrings (affinity.go:47-51) promise distinct behaviors:

- `PlanThenAct` — "generate plan, show as FYI, auto-execute"
- `PlanInParallel` — "generate plan, start execution in parallel"
- `JustAct` — "skip plan generation entirely"

Only the gating decision is wired. JustAct doesn't skip plan generation — `handleTask` still calls `a.Plan(...)` unconditionally before `shouldRequireApproval` even runs. Either:
- Wire the distinct behaviors (move the affinity decision *before* plan generation in `handleTask`), or
- Collapse the three tiers down to `Gate / NoGate` until the parallel-execute / skip-plan paths exist.

Right now the type promises a richer API than the implementation delivers, which is the kind of debt that bites later.

### 3. `a.lastCluster` is set but never read (agent.go:62, 663)

Declared on the struct, assigned in the auto-execute branch of `handleTask`, never read anywhere. The comment at agent.go:638-641 acknowledges THANKED/FIXED need a post-turn callback that doesn't exist yet. Until that callback lands, the field is dead — drop it from the struct now and reintroduce when the callback is wired, otherwise it just sits there as a maintenance hazard.

### 4. Missing test for the new "destructive always gates, even under `act`" safety property (router_test.go)

The old `TestShouldRequireApprovalNever` exercised the destructive-plan-under-never path (and asserted no gate). That assertion is now wrong (the safety change is the headline of `shouldRequireApproval`'s new docstring), so the test was rewritten to use a reversible plan — but no test was added in its place to lock in the *new* behavior. `TestShouldRequireApprovalAutoDestructiveStep` covers it for `auto` but not for an explicit `autonomy_floor=act`. Add:

```go
func TestShouldRequireApproval_ActStillGatesDestructive(t *testing.T) {
    a := New(config.AgentConfig{Name: "test", AutonomyFloor: "act"}, ...)
    plan := &core.Plan{Task: "clean up", Steps: []core.PlanStep{{ID: 1, Description: "rm -rf /tmp/foo"}}}
    if !a.shouldRequireApproval(ctx, plan) {
        t.Error("act floor must still gate destructive plans")
    }
}
```

### 5. `firstStepIsAskingTheUser` credits `OutcomeIgnored` against the cluster (agent.go:614-621)

When the planner's first step is "Ask the user for X", we credit IGNORED to the cluster of the original (vague) input. But `summarize this video` (no link) and `summarize this https://youtube.com/...` collapse to the *same* cluster (`summarize/youtube`) — and only the first form trips the ask-the-user shortcut. Crediting IGNORED here punishes the whole cluster for vagueness, when really we want to learn "the input was incomplete" not "this cluster doesn't need a plan."

Options: (a) don't credit anything in this branch — the cluster simply doesn't get a sample, (b) introduce a new `OutcomeIncomplete` outcome that doesn't pull score toward zero, or (c) only credit when the cluster has already accumulated samples (so we don't sabotage learning on fresh clusters).

### Nits (won't block)

- **affinity.go:13** — docstring says "After ~5 samples per cluster the agent can decide autonomously" but `minSamples = 3` at affinity.go:56. Pick one.
- **cluster.go:818** — the `.pdf` rule calls `strings.Fields(s)` twice and would panic on a zero-field input. `ClusterFor` guards empty inputs, so this is theoretical, but a `if fields := strings.Fields(s); len(fields) > 0 && strings.HasSuffix(fields[len(fields)-1], ".pdf")` would be more defensive.
- **affinity.go:30** — `affinity:cluster:` is documented as a reserved namespace, but nothing enforces it; a user typing `/remember affinity:cluster:abc = junk` could corrupt the store. Worth a `if strings.HasPrefix(key, AffinityKeyPrefix) return ErrReserved` guard in the user-facing memory store.

### Things I explicitly like

- Cluster/affinity/wiring separation. Easy to test each piece in isolation, and the test suite does exactly that.
- Pull-toward-extreme update rule with `[0.05, 0.95]` clamp is the right shape — clusters can always recover. The ring-buffer cap test pins the persistence contract nicely.
- Two-site migration (`config.fillDefaults` + `agent.New`) catching hand-built `AgentConfig` from tests/embedders — easy to forget, important to get right.
- `looksLikeStopAsking` collapses "flip the pref + approve the pending plan + credit the cluster" into one shot. Direct fix for bug #5/#11 — the "Yea" → "Yes" double-tap loop was exactly the kind of friction this should kill.
- Destructive-plans-always-gate, regardless of floor. Right safety default; the docstring is explicit about it.
- `firstStepIsAskingTheUser` short-circuit is a nice product fix for issue #17 — "I'll ask the user, then…" 5-step plans are wasteful theatre.

### Verdict

Re-requesting changes on items 1–5; nits won't block. Items 1, 2, 3 are the substantive code issues; 4 is a missing-test gap; 5 is a learning-signal correctness concern.

srvsngh99:
Thanks — addressed all five blockers + the two code nits in \`43138a3\`.

**1. Affinity narration on real threshold crossing** — \`ClusterAffinity\` gained a sticky \`WasAutoRun\` bit. \`AffinityStore.Update\` now returns \`(rec, flippedToAutoRun, err)\` so the caller narrates exactly once on the false→true transition, regardless of whether the cluster crossed at SampleCount=3 or after 7 approvals + a stray rejection. New \`TestAffinity_FlippedOnAutoRunCrossing\` exercises the path you flagged: one MODIFIED then nine APPROVEDs, one and only one flip.

**2. PlanDecision tiers wired honestly** — \`JustAct\` is now consumed at the top of \`handleTask\` *before* plan generation, short-circuiting to \`handleAnswer\`. No wasted planner round-trip on calibrated low-affinity clusters. PlanThenAct and PlanInParallel both reach \`shouldRequireApproval\` and currently behave identically (no gate); the docstring is now honest that the streaming/parallel UX is future work, so the type doesn't promise more than the implementation delivers.

**3. \`a.lastCluster\` deleted** — agreed, dead field, dropped from the struct + the assignment site.

**4. New \`TestShouldRequireApproval_ActStillGatesDestructive\`** — locks in the v0.1.1 safety property (no autonomy floor lets destructive plans through silently). \`AutonomyFloor=act\` + \`delete file unused.go\` step → expect \`true\`.

**5. \`firstStepIsAskingTheUser\` no longer credits OutcomeIgnored** — agreed with your reasoning: input was incomplete, not the cluster's fault. Punishing \`summarize/youtube\` for one missing-link request would sabotage learning. Now it just asks the question and moves on; the cluster gets no sample.

**Nit a (minSamples docstring)** — fixed (\`~3\` matches the constant).
**Nit b (pdf rule defensive guard)** — fixed (explicit \`len(fields) == 0\` check).
**Nit c (\`AffinityKeyPrefix\` reservation in memory store)** — leaving for follow-up since it touches a wider API surface and you flagged as non-blocking.

\`go build\`, \`go vet\`, \`go test ./...\` all green on \`43138a3\`. PTAL.

srvsngh99:
## Approval — PR #33, round 2 ✅

Pulled `43138a3`, re-ran `go build ./... && go vet ./... && go test ./...` — all green. Each round-1 item is cleanly addressed:

### 1. Narration fires on actual threshold crossing — fixed

`ClusterAffinity` now carries a sticky `WasAutoRun` bool (`affinity.go:58-71`) plus an `isAutoRun()` helper (`affinity.go:74-77`). `AffinityStore.Update` returns `(rec, flippedToAutoRun, err)` — `flippedToAutoRun` is `true` exactly once when `!wasAutoRunBefore && rec.WasAutoRun`. `creditOutcome` in `agent.go:507-521` propagates the bool; the caller in `handlePendingPlan:443-446` wraps `appendAffinityNarration` only when `flipped` is true. The new `TestAffinity_FlippedOnAutoRunCrossing` (`affinity_test.go:104-141`) exercises the exact regression I flagged — 1 MODIFIED + 9 APPROVED so the cluster crosses 0.7 well after `SampleCount==minSamples` — and asserts (a) exactly one flip within 10 updates, (b) `WasAutoRun` sticky, (c) no re-flip on a subsequent approval. Right semantics.

### 2. `PlanDecision` tiers wired honestly — fixed

`handleTask:622-633` now consults affinity *before* plan generation and short-circuits to `handleAnswer` when the decision is `DecisionJustAct`. No more wasted planner round-trip for clusters that have learned they don't need a plan. `shouldRequireApproval`'s docstring (`agent.go:786-797`) is now honest that `JustAct` is handled upstream and that `PlanThenAct` / `PlanInParallel` currently behave identically with the parallel/streaming UX called out as future work. Promises now match behavior.

### 3. Dead `lastCluster` field — removed

Struct field gone (`agent.go:62`), assignment gone (`agent.go:663`), THANKED/FIXED comment block removed. Clean.

### 4. Destructive + `act` floor gate test — added

`TestShouldRequireApproval_ActStillGatesDestructive` at `router_test.go:328-350` explicitly constructs `AutonomyFloor: "act"`, a single destructive step `"delete file unused.go"`, and asserts the gate fires. The safety property now has its own pinned assertion rather than relying on transitive coverage under `"auto"`.

### 5. `firstStepIsAskingTheUser` no longer credits IGNORED — fixed

The `a.creditOutcome(ctx, cluster, OutcomeIgnored)` call is gone from `agent.go:653-661`, and the comment now spells out why: "We deliberately do NOT credit an outcome here: the input was incomplete, not the cluster's fault." Stops sabotaging learning on otherwise-strong clusters.

### Nits

- `affinity.go:6-9` docstring now reads "After ~3 samples per cluster (minSamples)". Matches the constant.
- `cluster.go:67-77` pdf rule guards `len(fields) == 0` before indexing.
- The `affinity:cluster:` namespace-enforcement nit wasn't addressed, but it was non-blocking and stays in the follow-up bucket.

### Forward-looking observation (not blocking)

Once a cluster crosses `minSamples`, `shouldRequireApproval` returns `false` for every decision tier, so no more outcomes ever get credited — affinity learning effectively freezes at the crossing point. The only way out is `THANKED` / `FIXED`, which need the post-turn callback that doesn't exist yet (acknowledged in the commit message of round 2 and via the dead-field cleanup). Fine for v0.1.1 — flagging so the third PR in the cycle, or v0.1.2, can pick it up if it shows up in practice.

### Verdict

LGTM — approving. All five substantive items resolved, tests cover the right semantics (the `FlippedOnAutoRunCrossing` test in particular pins the exact regression I was worried about), and the API/docstrings are now honest about what's wired vs future. Ready to merge as the second of the three v0.1.1 PRs.

srvsngh99:
Thanks for the approval! Forward-looking observation noted — the post-crossing freeze on outcome credits is real and the THANKED/FIXED post-turn callback is the natural next step. Filing it as a v0.1.2 follow-up since it'd require new wiring (the user's next message after an auto-execute is the signal source, but ChatFromPlatform doesn't currently know about the prior auto-execute). Ready to merge whenever.

PR #34, opened 2026-05-15
No permission prompts by default; v0.1.2

srvsngh99:
## What

Two things, one patch release (**v0.1.2**):

### 1. Manual approval prompts are off by default (the main fix)
`shouldRequireApproval` had a legacy heuristic that, under the **default `act` floor**, still gated any non-destructive plan with **≥5 steps** or a "vague" task word (`everything`, `set up`, …). Because per-task-type affinity needs 3+ runs to calibrate, normal users hit that path constantly and got `Approve this plan? (yes/no)` on routine safe work — the top "why does it keep asking permission" complaint, and a direct contradiction of the documented v0.1.1 `act` contract ("non-destructive plans run without an approval gate").

Now: under `act`/`evolve` (the default), a **non-destructive plan never gates**. Step count, vagueness, and un-calibrated clusters no longer reintroduce a prompt.

**Unchanged safety:** destructive plans (`delete`, `git push`, `deploy`, `rm`, `drop`, …) still always gate regardless of floor. `autonomy_floor: suggest`/`observe` still gate on demand for users who want it.

The two tests that had locked in the old (buggy) behaviour are rewritten to assert the corrected contract.

### 2. v0.1.1 version regression fixed
The v0.1.1 release shipped its CHANGELOG and merged code but **never bumped `core.Version`** (still reported `0.1.0`) or the README badge. Both now report `0.1.2`.

## Test

`go build ./...`, `go vet ./...`, `go test ./...` — all green. Binary reports `minikrill version 0.1.2`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,17 @@
 All notable changes to Mini Krill will be documented in this file.
 Format based on [Keep a Changelog](https://keepachangelog.com/).
 
+## [0.1.2] - 2026-05-16
+
+No more nagging. The agent stops asking for manual permission on routine work.
+
+### Autonomy
+- Manual approval prompts are **off by default**. Under the default `act`/`evolve` autonomy floor, a non-destructive plan never gates — step count, task vagueness, and an un-calibrated affinity cluster are no longer reasons to ask "Approve this plan? (yes/no)". This was the top "why does it keep asking permission" complaint and contradicted the documented v0.1.1 `act` contract. Users who want a gate set `autonomy_floor: suggest`.
+- Unchanged safety: destructive plans (delete, `git push`, deploy, `rm`, drop, …) still always gate, regardless of floor. `autonomy_floor: suggest`/`observe` still gate on demand.
+
+### Fixed
+- Version constant and README badge corrected — the v0.1.1 release shipped its CHANGELOG and merged code but never bumped `core.Version` (still reported `0.1.0`) or the README badge. Both now report `0.1.2`.
+
 ## [0.1.1] - 2026-05-15
 
 Trust, autonomy, and identity. The agent now learns which task types you want planned, acts decisively on the rest, and is no longer hard-wired as a krill.
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@
 
 Built by [Sourav Singh](https://souravailabs.ai/about/) / [Sourav AI Labs](https://souravailabs.ai)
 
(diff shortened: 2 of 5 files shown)


srvsngh99:
## Review — PR #34 (round 1) — LGTM

**Checked at** head `2f399e8`. `go build ./...`, `go vet ./...`, `go test ./...` all green; `go test ./internal/agent/ -run TestShouldRequireApproval -count=1` re-run fresh (uncached) — all pass.

### Verification of the change

1. **`shouldRequireApproval` (agent.go:854-865)** — under `act`/`evolve` the function now returns `false` unconditionally for non-destructive plans. `planIsDestructive` still short-circuits to `true` at the top (845-846), and `observe`/`suggest` still return `true` (850-853). This matches the documented v0.1.1 `act` contract. Correct.

2. **"Unreachable in practice" claim is accurate.** Verified the defensive legacy fallback (868-883) cannot be hit from real config: `config.go:317-321` validates `AutonomyFloor` and coerces any unrecognised value to `act` on load, and `New()` (agent.go:76-85) normalises an empty floor. The only way into the fallback is a programmatic caller setting an invalid *non-empty* floor directly — genuinely defensive-only, so the comment is honest, not aspirational.

3. **No dead code.** The affinity store is no longer consulted here, but it's still used upstream at agent.go:682 for the `act`/`evolve` JustAct path, so calibration still happens and `a.affinity` is not orphaned.

4. **Test rewrites assert the corrected contract, not the bug.** `TestShouldRequireApprovalAutoLargePlan` (6-step reversible plan → no gate) and `TestShouldRequireApprovalAutoVagueTask` ("improve everything…" → no gate) now assert the right behaviour; the latter's added `PlanApproval:"always"` → `suggest` sub-case correctly proves the gate is still available on demand. `TestShouldRequireApproval_ActStillGatesDestructive` and `AutoDestructiveStep` still pass, confirming destructive plans gate regardless of floor.

5. **Version bump complete and consistent.** `core.Version` `0.1.0`→`0.1.2`, README badge `0.1.0`→`0.1.2`, CHANGELOG entry added. Grepped the tree for stale `0.1.0`/`0.1.1` strings — only remaining hit is an accurate "(the default floor since v0.1.1)" code comment. Nothing left reporting an old version.

### Nits

None blocking. (No follow-ups to track.)

### Verdict

**LGTM — approving.** Logic change faithfully implements the documented `act` contract, safety on destructive plans is preserved, the "unreachable fallback" claim is backed by config-layer validation, tests assert the fixed behaviour, and the v0.1.1 version regression is fully closed.

PR #16, opened 2026-04-30
Fix codex CLI flags and improve TUI usability

srvsngh99:
## Summary
- **Fix codex chat**: Replace invalid `--ask-for-approval never` and `--sandbox read-only` flags with `--full-auto` for codex-cli 0.125.0 compatibility. This was causing `unexpected argument` errors on every LLM call across all interfaces (CLI, TUI, Telegram, Discord). Referenced deepkrill's working implementation.
- **Add log copy**: Press `c` on the Logs tab to copy all visible log content to clipboard. Shows a brief "✓ copied!" confirmation that auto-clears.
- **Add keybinding hints**: Footer now shows `Tab: switch tabs  ?: help  q: quit` so users can discover tab navigation without needing to find the Help tab first.

## Test plan
- [x] `go build ./cmd/minikrill/` passes
- [x] `go test ./...` passes (all packages)
- [x] Smoke tested `codex exec --full-auto --skip-git-repo-check --ephemeral --color never -` with stdin prompt — works correctly
- [ ] Verify codex chat works in TUI (`minikrill chat`)
- [ ] Verify `c` key copies logs on Logs tab
- [ ] Verify footer hints are visible on all tabs

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #35, opened 2026-05-16
Closed-loop learning: post-turn THANKED/FIXED, #16, reservation guard

srvsngh99:
First PR of the v0.1.3 cycle (per ~/.claude/plans/mini-krill-v012-cycle.md). Scoped to the **closed-loop learning core** — the load-bearing deferred item from the PR #33 review.

## Why

The affinity learner was one-directional. The approval path credits APPROVED/MODIFIED/REJECTED, but an **auto-run produces no verdict**, so once a cluster crossed `minSamples` its score could only ever rise. A cluster that started producing bad output would stay auto-run forever — no path back to "plan first".

## What

1. **Post-turn THANKED/FIXED callback.** New `KrillAgent.lastAutoRun` records the cluster whenever a turn executes un-gated (auto-exec, affinity JustAct skip, background submit). The user's *next* message is the implicit verdict, classified at the top of `ChatFromPlatform`:
   - correction against the auto-run's response → **FIXED** (planning would have helped, `+0.15`)
   - anything else — praise, follow-up, new topic, slash command → **THANKED** (no complaint = plan unneeded, `-0.10`)

   Asymmetric on purpose: a real correction pulls back harder than a neutral follow-up nudges down. Restores symmetric affinity drift.

2. **#16 correction-trigger asymmetry.** `looksLikeCorrection` caught `"don't"` but not `doesn't`/`didn't`/`isn't`/`wasn't`/`won't`/`can't`/… so genuine corrections were under-counted and never credited FIXED. Extended the contraction list (word-bounded via existing `containsWord`) and added curly-apostrophe (’) normalisation for phone keyboards. The redirection guard (forward-verb → not a correction) is unchanged.

3. **AffinityKeyPrefix reservation guard.** `FileMemory.Store` now refuses writes under `affinity:cluster:` unless `Source == core.AffinitySource`, so `/remember affinity:cluster:... = junk` can't corrupt a learned record. Introduces shared `core.ReservedAffinityPrefix` / `core.AffinitySource` / `core.ErrReservedKey` so the writer and the guard can't drift apart.

## Scope note

Episodic memory (#29/D6) and user-state (#37) — the other two PR A items in the plan — are split into a focused follow-up PR to keep this one reviewable; these three are tightly cohesive (they all close the affinity loop) and self-contained.

## Test

New: contraction asymmetry + redirection guard, post-turn outcome classification, `lastAutoRun` single-consume, reservation guard (refuse / exempt / unaffected). `go build ./...`, `go vet ./...`, `go test ./...` all green.

### DIFF
--- a/internal/agent/affinity.go
+++ b/internal/agent/affinity.go
@@ -20,8 +20,10 @@ import (
 )
 
 // AffinityKeyPrefix reserves a memory key namespace so listing/searching the
-// regular memory store skips affinity entries.
-const AffinityKeyPrefix = "affinity:cluster:"
+// regular memory store skips affinity entries. It aliases the shared core
+// constant so the brain-side reservation guard and this writer can never
+// drift apart.
+const AffinityKeyPrefix = core.ReservedAffinityPrefix
 
 // AffinityOutcome is the discrete event types the store understands.
 type AffinityOutcome string
@@ -131,7 +133,7 @@ func (s *AffinityStore) Put(ctx context.Context, rec ClusterAffinity) error {
 		Value:      string(data),
 		Tags:       []string{"affinity", "system"},
 		Scope:      "system",
-		Source:     "affinity-store",
+		Source:     core.AffinitySource,
 		CreatedAt:  rec.LastUpdate,
 		AccessedAt: rec.LastUpdate,
 	})
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -61,6 +61,7 @@ type KrillAgent struct {
 	turnFetches       *turnFetchLog  // per-turn provenance ledger; reset on each ChatFromPlatform call
 	affinity          *AffinityStore // task-type plan-affinity learner
 	pendingClstr      TaskCluster    // cluster while a plan is awaiting approval
+	lastAutoRun       TaskCluster    // cluster of the most recent auto-executed (un-gated) turn, awaiting a post-turn THANKED/FIXED verdict from the user's next message
 	cachedEmojiStyle  string         // cached read of cfg.Brain.EmojiStyle; refreshed on /emoji
 	cachedOverlayDirs []string       // cached overlay directives; invalidated on /persona writes
 	cachedOverlayInit bool           // true once cachedOverlayDirs has been loaded at least once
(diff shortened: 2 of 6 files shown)


srvsngh99:
## Review — PR #35 (round 1) — LGTM

**Checked at** head `fa2b562`, reviewed in an isolated git worktree (the repo is shared by concurrent sessions). `go build ./...`, `go vet ./...`, full `go test ./...`, and **`go test -race ./internal/agent/ ./internal/brain/`** all green. The four new tests pass individually:

```
--- PASS: TestLooksLikeCorrection_ContractionAsymmetry
--- PASS: TestPostTurnOutcome
--- PASS: TestCreditPostTurn_ConsumesLastAutoRun
--- PASS: TestStore_ReservedAffinityPrefix
```

This closes the affinity loop cleanly. I verified the load-bearing correctness concerns rather than taking the description on trust:

1. **No data race on `lastAutoRun` / `history`** — `ChatFromPlatform` takes `a.mu.Lock()` / `defer a.mu.Unlock()` at the very top (`agent.go:213-214`), and both the write (`handleTask` → `a.lastAutoRun = cluster`, `agent.go:732`/`771`) and the read+consume (`creditPostTurn`, `agent.go:229`) happen inside that critical section. The async `go a.creditOutcome(...)` only receives `cluster`/`outcome` by value and touches the separately-synchronised affinity store — it never aliases `a.*`. `-race` on the agent package confirms this empirically.
2. **`lastAssistantBefore` is bounds-safe** — `creditPostTurn` passes `len(a.history)-1`, which is `-1` for an empty history; `lastAssistantBefore` (`agent.go:1570-1580`) guards `before >= len` and the `for i := before; i >= 0` loop simply no-ops at `i == -1`, returning `""`. Empty prev → `looksLikeCorrection` fails its ≥30-char guard → THANKED. Safe default, no panic.
3. **Reservation guard is correct and drift-proof** — `memory.go:63-66` refuses `ReservedAffinityPrefix` writes unless `Source == AffinitySource`; `affinity.go:136` now stamps `core.AffinitySource` and `AffinityKeyPrefix` aliases `core.ReservedAffinityPrefix`, so the writer and the guard share a single source of truth and cannot diverge. `fmt`/`strings` already imported. Test covers refuse / exempt / unaffected.
4. **Weights match the stated asymmetry** — `affinity.go:168-169` THANKED `score -= 0.10*score`, FIXED `score += 0.15*(1-score)`. A real correction pulls back harder than a neutral follow-up nudges down, restoring the symmetric drift the description promises.

### Nits (won't block — do not patch into this PR)

1. **Heuristic false-positive on polite dismissal** — the expanded contraction list now matches phrases like *"that won't be necessary, thanks"* as a correction → FIXED rather than THANKED. It's mitigated by the ≤8-word + forward-verb + prev-length guards and the bounded `+0.15*(1-score)` step, and false-positives are inherent to a keyword heuristic, so this is acceptable — worth only a one-line code comment acknowledging the known edge, not a logic change.
2. **Slash command as implicit verdict** — `creditPostTurn` runs before `handleProviderCommand`, so a reflexive `/model` after an auto-run still credits THANKED. The PR explicitly intends this ("even a slash command → THANKED") and the −0.10 nudge is weak, so it's a documented design choice; flagging only so it's a conscious one.

### Things I like

- The async-credit-by-value pattern mirrors the existing `recordFeedback` path — consistent and lock-safe by construction.
- `postTurnOutcome` extracted as a pure function makes the classification unit-testable without the goroutine; the test exercises praise / neutral-follow-up / correction.
- Curly-apostrophe normalisation inside `looksLikeCorrection` (not at call sites) means every caller benefits and the contraction list stays canonical.
- Shared `core.ReservedAffinityPrefix` / `AffinitySource` / `ErrReservedKey` is exactly the right fix for writer/guard drift, and the test asserts all three guard branches.
- Scope discipline: splitting episodic memory / user-state into a follow-up keeps this cohesive and reviewable.

### Verdict

**LGTM — approving.** Build/vet/test/-race green, the affinity loop is now symmetric, the concurrency and bounds edges are safe on inspection, and the reservation guard is drift-proof. Nits are non-blocking and explicitly should not be patched into this approved PR. Not merging (separate explicit step).

PR #36, opened 2026-05-16
Episodic memory + user-state (#29/D6, #37)

srvsngh99:
Second PR of the v0.1.3 cycle. Follows #35 (now merged); rebased onto current `main`. Completes the remaining two PR A items from the plan, split out so #35 stayed reviewable.

## Episodic memory (#29 / D6) — `internal/brain/episodic.go`
On a **>30 min activity gap** the agent asynchronously summarises the prior session into one episode `MemoryEntry` (tag `episode`, `Source: reflection`). The most recent episode **<7 days old** is injected into chat/answer context as an explicitly **point-in-time, read-only** line, so the agent resumes context instead of starting cold — without mistaking a stale summary for live state (same discipline long-term memory uses).

- Dedupes (25 min window), no-ops on thin sessions (<4 turns), safe no-op when any dependency is nil.
- Wired through `*brain.KrillBrain` + a **local `episodicBrain` interface** in agent (type assertion), so the agent still depends only on `core` — never the `brain` package — and mock brains transparently no-op.

## User-state (#37) — `internal/agent/userstate.go`
Models the user as a stateful agent: focus area, last task cluster, light keyword mood, last-seen. Persisted to a normal memory key, refreshed async each turn, injected read-only + point-in-time into chat context. Behavioural gating is deliberately **not** included — read-only context only, to avoid risk; routing already handles greeting-vs-task.

## Test
Episodic: consolidate / dedupe / thin-session / disabled-no-op, `Latest` maxAge filter + newest-pick. User-state: `lightMood`, round-trip + framing, nil-memory safety. `go build`/`vet`/`test ./...` all green.

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -62,6 +62,7 @@ type KrillAgent struct {
 	affinity          *AffinityStore // task-type plan-affinity learner
 	pendingClstr      TaskCluster    // cluster while a plan is awaiting approval
 	lastAutoRun       TaskCluster    // cluster of the most recent auto-executed (un-gated) turn, awaiting a post-turn THANKED/FIXED verdict from the user's next message
+	lastTurnAt        time.Time      // wall-clock of the previous turn; a large gap marks a session boundary for episodic consolidation
 	cachedEmojiStyle  string         // cached read of cfg.Brain.EmojiStyle; refreshed on /emoji
 	cachedOverlayDirs []string       // cached overlay directives; invalidated on /persona writes
 	cachedOverlayInit bool           // true once cachedOverlayDirs has been loaded at least once
@@ -227,6 +228,27 @@ func (a *KrillAgent) ChatFromPlatform(ctx context.Context, platform, chatID, inp
 	// a cluster could graduate to auto-run but never fall back, so a cluster
 	// that started producing bad output stayed auto-run forever.
 	a.creditPostTurn(input)
+
+	// Session-boundary episodic consolidation (#29 / D6): a long gap since the
+	// previous turn means the prior session ended. Summarise it asynchronously
+	// now so the *next* greeting can reference it. The consolidator dedupes
+	// and no-ops on thin sessions, so an over-eager trigger is harmless.
+	now := time.Now()
+	if !a.lastTurnAt.IsZero() && now.Sub(a.lastTurnAt) > sessionGap {
+		if eb, ok := a.brain.(episodicBrain); ok {
+			ch := a.channel
+			go func() { _ = eb.ConsolidateEpisode(context.Background(), ch) }()
+		}
+	}
+	a.lastTurnAt = now
+
+	// Model the user as a stateful agent (#37): refresh focus / mood /
+	// last-seen from this message. Async — the file write never blocks the
+	// reply, and a slightly stale read next turn is acceptable for context.
+	if mem := a.brain.Memory(); mem != nil {
+		in := input
+		go updateUserState(context.Background(), mem, in)
+	}
 
 	if response, handled := a.handleProviderCommand(input); handled {
 		a.saveTurn("user", input)
@@ -578,6 +600,19 @@ func (a *KrillAgent) creditOutcome(ctx context.Context, cluster TaskCluster, out
 		return false
 	}
 	return flipped
+}
+
+// sessionGap is how long without a turn marks the previous session as ended,
+// triggering an episodic summary of it.
+const sessionGap = 30 * time.Minute
+
+// episodicBrain is the optional capability the agent uses for cross-session
+// memory (#29 / D6). It's a local interface satisfied by *brain.KrillBrain
+// via a type assertion, so the agent keeps depending only on core — never on
+// the brain package — and test mock brains transparently no-op.
+type episodicBrain interface {
+	ConsolidateEpisode(ctx context.Context, channel string) error
+	LatestEpisode(ctx context.Context, maxAge time.Duration) (string, error)
 }
 
 // creditPostTurn observes the user's first message after an auto-executed
@@ -982,6 +1017,22 @@ func (a *KrillAgent) handleChat(ctx context.Context) (string, error) {
 		enriched = insertBeforeLast(enriched, memoryMsg)
 	}
 
+	// Cross-session continuity (#29 / D6): surface the most recent episode if
+	// it's fresh enough to still be relevant. Capped at 7 days so the agent
+	// references "last session" but never a stale month-old summary.
+	if eb, ok := a.brain.(episodicBrain); ok {
+		if ep, _ := eb.LatestEpisode(ctx, 7*24*time.Hour); ep != "" {
+			enriched = insertBeforeLast(enriched, core.Message{Role: "system", Content: ep})
+		}
+	}
+
+	// User-as-agent state (#37): focus / mood / recency, read-only context.
+	if mem := a.brain.Memory(); mem != nil {
+		if line := loadUserState(ctx, mem).contextLine(); line != "" {
+			enriched = insertBeforeLast(enriched, core.Message{Role: "system", Content: line})
+		}
+	}
+
 	if styleDirective := buildStyleDirective(ctx, a.brain.Memory()); styleDirective != "" {
 		styleMsg := core.Message{Role: "system", Content: styleDirective}
 		enriched = insertBeforeLast(enriched, styleMsg)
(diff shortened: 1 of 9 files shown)


srvsngh99:
## Review — PR #36 (round 1)

**Checked at** head `99d3408`, reviewed in an isolated git worktree. `go build ./...`, `go vet ./...`, full `go test ./...`, and `go test -race ./internal/agent/ ./internal/brain/` all green; CI is green too. The design (local `episodicBrain` interface keeping agent→core-only, nil-safe no-ops, point-in-time framing, mutex-safe `ConversationStore` access) is sound. But green CI does not exercise the issue below — the fixtures only ever create ≤2 episodes.

### Substantive issues

1. **`latestEntry` returns the *oldest* 50 episodes, so cross-session continuity silently dies after ~50 episodes — and the dedupe guard fails with it.** `episodic.go:493` calls `e.mem.Search(ctx, "episode_", 50)`. `FileMemory.Search` (`memory.go:138-148`) iterates `listAll()` and `break`s once it has `limit` matches — it takes the *first* 50 in iteration order, with no recency sort. `listAll()` is `os.ReadDir(m.dir)` (`memory.go:323`), which returns filenames in lexical order. Episode keys are `episode_<unixMilli>` — a fixed-width, monotonically increasing integer — so lexical order is chronological **ascending**. Therefore `Search("episode_", 50)` yields the 50 *oldest* episodes, and `latestEntry` picks "the newest of the oldest 50". Consequences once a user accumulates >50 episodes (weeks of normal use):
   - `Latest()` (`episodic.go:481`) operates on a stale subset; once those 50 oldest all age past its 7-day `maxAge` it returns `""` permanently — the feature quietly stops with no error.
   - `Consolidate`'s dedupe (`episodic.go:421`, `time.Since(recent.CreatedAt) < episodeDedupeWindow`) also calls `latestEntry`; once it returns an episode older than 25 min, the dedupe check is always false, so the guard stops suppressing and every >30-min-gap turn can mint duplicates — *accelerating* the accumulation that triggered the bug.

   **Fix:** don't depend on `Search`+manual-max for "newest". Options: (a) a recency-ordered lookup (sort episode entries by `CreatedAt` desc before truncating), (b) scan *all* episode-tagged entries (`Search` with `limit=0`, then pick `max(CreatedAt)`) — bounded only if (c) is also done, or (c) keep a single rolling `episode:latest` pointer key the consolidator overwrites, plus prune old `episode_*` entries (see below). (c) is the most robust and also bounds growth.

2. **The episode namespace grows unbounded, and the consolidator never fires for the most common session boundary (process restart).** Two coupled problems: (i) `FileMemory` takes `maxMemories` but `Store` (`memory.go:59-125`) never enforces it — nothing prunes `episode_*`, so it grows without limit, which is also what makes issue 1 inevitable. (ii) The trigger in `ChatFromPlatform` (`agent.go:222`) requires `!a.lastTurnAt.IsZero()`, but `lastTurnAt` (`agent.go:65`) is an in-memory field, zero on every process start. The headline use case — a CLI session that ends by the process exiting and resumes in a *new* process later — therefore never consolidates; only a single long-lived process (e.g. the Telegram daemon) seeing a >30-min in-process idle gap does. The feature under-delivers on its stated "resume across sessions" premise for exactly the restart case. **Fix:** derive the gap from a persisted signal (e.g. the last turn's timestamp in `ConversationStore`) rather than an in-memory field, and add episode pruning/TTL so issue 1 can't recur.

### Nits (won't block — do not patch into an approved PR later)

1. The ~14-line episode + user-state injection block is duplicated verbatim in `handleChat` (`agent.go:1020-1035`) and `handleAnswer` (`agent.go:1377-1392`). Extract one helper — duplicated context-assembly is precisely the writer/guard-drift risk #35 just fixed elsewhere.
2. `Latest` renders age via `time.Since(...).Round(time.Hour)`, so any episode <30 min old prints "~0s ago". Use a coarser human format or `Round(time.Minute)`.
3. `updateUserState` does a load-modify-save across two independently-locked `FileMemory` ops from an async goroutine, so two near-simultaneous turns can lose an update. The PR explicitly tolerates staleness so this is acceptable — worth a one-line code comment making that a conscious choice.

### Things I like

- `episodicBrain` as a local interface satisfied by `*KrillBrain` via type assertion — agent keeps depending only on `core`, mock brains transparently no-op. Clean.
- Nil-dependency-disables-everything (`Episodic.enabled()`, nil-receiver no-op) is tested and keeps headless/test paths simple.
- Point-in-time, read-only framing on both episode and user-state lines, consistent with the long-term-memory discipline; behavioural gating deliberately excluded for #37 is the right scope call.
- `ConversationStore.SaveTurn`/`LoadRecent` are mutex-protected, so the async consolidate goroutine racing the main turn's `saveTurn` is safe — verified, not assumed.
- Dedupe window + thin-session guard + 80-turn cap are sensible guardrails (modulo issue 1 undermining the dedupe).

### Verdict

**Re-requesting changes** for **item 1** (newest-episode lookup is actually an oldest-episode lookup → feature silently dies and dedupe breaks after ~50 episodes; blocking — it defeats the PR's purpose in normal use) and **item 2** (unbounded growth + never fires across process restarts). Nits won't block. CI being green is expected here — the fixtures never reach the >50-episode regime where item 1 bites, which is exactly why this needs the structural fix rather than a test tweak.

srvsngh99:
Round-1 review addressed in 5459b51.

**Item 1 (blocking) — newest-episode lookup was an oldest-episode lookup.** Replaced the `Search("episode_", 50)`+manual-max scan with a single rolling `episode:latest` pointer the consolidator overwrites; `latestEntry` is now an O(1) exact-key `Recall`, so "newest" is correct regardless of history size and the dedupe guard can't degrade. (Option (c) from the review.)

**Item 2 — unbounded growth + never fires across process restart.** (i) Added `pruneHistory`: after each consolidate, episode_* is bounded to the newest `maxEpisodeHistory` (50); the pointer is a distinct key and never pruned. (ii) The trigger no longer keys off in-memory `lastTurnAt` (zero every process start). It now measures the gap against a new persisted `ConversationStore.LastActivity(channel)`, so a session that ended by the process exiting consolidates when a later process resumes — the headline case. `lastTurnAt` removed.

**Nits:** extracted `injectSessionContext` (was duplicated verbatim in `handleChat`/`handleAnswer`); `humanizeAge` replaces `Round(time.Hour)` so a fresh episode no longer prints `~0s ago`; added a comment marking the userstate load-modify-save staleness as a deliberate trade-off.

**Tests:** `TestEpisodic_LatestSurvivesManyEpisodes` seeds 60 stale episodes and asserts Latest returns the fresh one + history is pruned (this is the regime the old fixtures never reached); plus `TestLastActivity`, `TestHumanizeAge`. `go build`/`vet`/`test ./...` and `go test -race ./internal/agent/ ./internal/brain/` all green.

srvsngh99:
## Review — PR #36 (round 2) — LGTM

**Checked at** head `5459b51` ("rolling episode pointer + prune + restart-safe trigger"), reviewed in an isolated worktree. `go build ./...`, `go vet ./...`, full `go test ./...`, and `go test -race ./internal/agent/ ./internal/brain/` all green; **CI green** (6 builds + Lint + Test). Both round-1 items resolved, all 3 nits addressed, no new issues.

### Round-1 item 1 (BLOCKING — newest-episode lookup was an oldest lookup) — ✅ resolved

`episodic.go:177-183` — `latestEntry` is now an O(1) exact-key `e.mem.Recall(episodeLatestKey)` ("episode:latest"), a rolling pointer the consolidator overwrites on every `Consolidate` (`episodic.go:118-121`). "Newest" is now correct independent of history size, and since the dedupe guard (`episodic.go:68`) goes through the same `latestEntry`, it can no longer degrade. The code comment correctly documents *why* a `Search` scan was wrong (oldest-first). `TestEpisodic_LatestSurvivesManyEpisodes` seeds 60 stale episodes and asserts `Latest` still returns the fresh one — exactly the >50 regime the round-1 fixtures never reached, and it passes. This is option (c) from the review, the most robust choice.

### Round-1 item 2 (unbounded growth + never fires across restart) — ✅ resolved

- **Pruning:** `pruneHistory` (`episodic.go:134-155`) runs after each consolidate (`:125`), filters to `episode_`-prefixed tagged entries (the `episode:latest` pointer is excluded — distinct key, verified), `sort.Slice` **newest-first** by `CreatedAt`, then `Forget`s `hist[maxEpisodeHistory:]` (everything past the newest 50). The sort direction is correct — it keeps newest and deletes oldest, not the inverse. Namespace is now bounded.
- **Restart-safe trigger:** `agent.go:240` now gates on the persisted `store.LastActivity(ch)` instead of in-memory `lastTurnAt` (removed entirely). `ConversationStore.LastActivity` (`conversations.go:134-135`) is `s.mu.Lock()`-guarded, so it's concurrency-safe with `SaveTurn`. A session that ended by the process exiting now consolidates when a later process resumes — the headline case the old code missed. `TestLastActivity` covers it.

### Round-1 nits — all addressed

- `injectSessionContext` (`agent.go:625`) extracted; called once each from `handleChat` (`:1041`) and `handleAnswer` (`:1384`) — the verbatim duplication is gone.
- `humanizeAge` (`episodic.go:187`) replaces `Round(time.Hour)`; a sub-30-min episode no longer prints "~0s ago". `TestHumanizeAge` covers it.
- `userstate.go:97-98` now explicitly documents the load-modify-save staleness as a deliberate, accepted trade-off (read-only context, not behavioural state).

### Verdict

**LGTM — approving.** Item 1's fix is correct by construction and proven by a test in the exact regime that previously broke; item 2 is bounded and restart-safe with the right sort direction; nits cleared. Build/vet/test/-race and CI all green. Not merging (separate explicit step).

PR #37, opened 2026-05-16
Owner-gated shared spaces: owner auth + crosspost-leak scrub

srvsngh99:
Third PR of the v0.1.3 cycle. Implements the **single-owner safety keystone** plus the small crosspost-leak fix. Owner identification via explicit config field (your chosen option).

## Owner authorization (#24, the keystone)

New `telegram.owner_id` / `discord.owner_id` config fields. When set, **only the owner drives the bot**. A non-owner in a shared group/server is a *bystander*:
- replied to with a fixed decline **only** if they directly @mention or reply-to the bot,
- their message **never reaches the agent** — so a stranger can't trigger tasks, memory writes, or destructive actions.

This is the safety counterpart to "no approval prompts by default" (#34): un-gated auto-execution is only safe if a stranger can't drive it. Enforced at the Telegram/Discord boundary with a pure `isOwner` classifier, so no authority flag has to thread through the agent.

**Backward compatible:** `owner_id` unset = current behaviour, with a one-time startup WARN recommending it for shared-group safety. Existing solo setups keep working untouched.

## #22 — crosspost-leak scrub

`scrubCrosspostLiterals` strips any residual `[CROSSPOST:..]` / `[/CROSSPOST]` literal the directive parser leaves behind on a malformed/unclosed directive, so the raw token never leaks into the user's chat.

## Deliberately deferred (with reasons)

These plan items are **not** in this PR — rushing them in a shared multi-session repo would be irresponsible, and two are riskier than the audit implied:

- **Per-space context scoping (#24 storage side)** — the plan's own Risk #2: changing the default `unified` channel breaks existing cross-interface continuity; needs a migration + `/recall across` opt-in. Wants its own focused PR.
- **E3.1 reply-to capture** — needs a new persisted `platform_msgid.jsonl` + GC; medium build, no safety urgency.
- **#9 Telegram backoff** — the audit described a "hardcoded 3s retry"; the code actually uses the library's `GetUpdatesChan`, so real backoff means replacing the live polling loop — an invasive refactor, not a contained change.

These are re-parked in `~/.claude/plans/mini-krill-v012-cycle.md`.

## Test
`isOwner` set/unset (Telegram + Discord), crosspost scrub (clean / unclosed-header / stray-close / unterminated). `go build`/`vet`/`test ./...` all green.

### DIFF
--- a/internal/chat/discord.go
+++ b/internal/chat/discord.go
@@ -141,6 +141,14 @@ func (d *DiscordBot) onMessageCreate(ctx context.Context) func(s *discordgo.Sess
 						break
 					}
 				}
+				// Parity with Telegram's isReplyToMe: a reply to the bot's
+				// own message counts as directly addressing it, even without
+				// an explicit @mention.
+				if !addressed && m.ReferencedMessage != nil &&
+					m.ReferencedMessage.Author != nil &&
+					m.ReferencedMessage.Author.ID == s.State.User.ID {
+					addressed = true
+				}
 				if addressed {
 					log.Info("bystander addressed the bot; declining (owner-gated)", "user_id", m.Author.ID, "username", m.Author.Username)
 					if _, err := s.ChannelMessageSend(m.ChannelID, discordBystanderDecline); err != nil {
--- a/internal/chat/discord.go
+++ b/internal/chat/discord.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"math/rand"
 	"strings"
+	"sync"
 
 	"github.com/bwmarrin/discordgo"
 
@@ -18,11 +19,20 @@ const discordMaxLen = 2000
 
 // DiscordBot implements core.ChatBot for Discord.
 type DiscordBot struct {
-	session *discordgo.Session
-	handler core.ChatHandler
-	cfg     config.DiscordConfig
-	done    chan struct{}
+	session   *discordgo.Session
+	handler   core.ChatHandler
+	cfg       config.DiscordConfig
+	done      chan struct{}
+	ownerWarn sync.Once // emits the "owner_id unset" hint at most once
 }
+
+// isOwner reports whether id is the configured single owner. Unset ("") →
+// always false → owner-gating off (legacy behaviour).
+func (d *DiscordBot) isOwner(id string) bool {
+	return d.cfg.OwnerID != "" && id == d.cfg.OwnerID
+}
+
+const discordBystanderDecline = "Hey! 🦐 I only take instructions from my owner, so I can't run tasks or remember things for anyone else here."
 
 // NewDiscordBot creates a DiscordBot with the provided config and handler.
 // The session is created but not opened - call Start to connect.
@@ -111,6 +121,36 @@ func (d *DiscordBot) onMessageCreate(ctx context.Context) func(s *discordgo.Sess
 		// Never reply to ourselves.
 		if m.Author.ID == s.State.User.ID {
 			return
+		}
+
+		// Owner authorization (single-owner model). When owner_id is set,
+		// non-owners are bystanders: replied to with a fixed decline only if
+		// they directly @mention the bot, and never routed to the agent — no
+		// tasks, memory writes, or destructive actions from a bystander. Unset
+		// → legacy behaviour with a one-time hint.
+		if !d.isOwner(m.Author.ID) {
+			if d.cfg.OwnerID == "" {
+				d.ownerWarn.Do(func() {
+					log.Warn("discord.owner_id is unset — anyone in a shared server can drive the bot; set discord.owner_id to your Discord user ID for single-owner safety")
+				})
+			} else {
+				addressed := m.GuildID == ""
+				for _, u := range m.Mentions {
+					if u.ID == s.State.User.ID {
+						addressed = true
+						break
+					}
+				}
+				if addressed {
+					log.Info("bystander addressed the bot; declining (owner-gated)", "user_id", m.Author.ID, "username", m.Author.Username)
+					if _, err := s.ChannelMessageSend(m.ChannelID, discordBystanderDecline); err != nil {
+						log.Debug("failed to send bystander decline", "error", err)
+					}
+				} else {
+					log.Debug("bystander message ignored (owner-gated)", "user_id", m.Author.ID)
+				}
+				return
+			}
 		}
 
 		// Guild/channel filter: if configured, only respond in matching contexts.
--- a/internal/chat/owner_gating_test.go
+++ b/internal/chat/owner_gating_test.go
@@ -1,0 +1,68 @@
+package chat
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/srvsngh99/mini-krill/internal/config"
+)
+
+func TestTelegramIsOwner(t *testing.T) {
+	// Unset → owner-gating off: nobody is "the owner", so the caller falls
+	// through to legacy behaviour.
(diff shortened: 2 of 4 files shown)


srvsngh99:
## Review — PR #37 (round 1)

**Checked at** head `de345e8`, reviewed in an isolated git worktree. `go build ./...`, `go vet ./...`, full `go test ./...`, `go test -race ./internal/chat/` all green; CI green. The boundary-classifier design (pure `isOwner`, no authority flag threaded through the agent), `sync.Once` one-time warn, and backward-compatible unset behaviour are all right, and they correctly implement the single-owner model. But there is one security gap the tests don't exercise.

### Substantive issues

1. **Telegram owner gate is allow-by-default when `msg.From == nil` — an anonymous group admin bypasses the keystone entirely.** `telegram.go:186` gates with `if msg.From != nil && !t.isOwner(msg.From.ID)`. When `OwnerID` is set but `msg.From == nil`, the whole authorization block is skipped and execution falls straight through to `msg.IsCommand()` → `handleCommand` → the agent. On Telegram `From == nil` is not exotic: it is exactly the case for **anonymous group admins** (any admin can toggle "remain anonymous"; sender becomes `SenderChat`, `From` is nil), channel-as-sender posts, and automatic forwards from a linked channel into a discussion group. So in the precise threat model this PR exists to close — "a stranger in a shared group can't drive the bot, trigger tasks, memory writes, or destructive actions" — a group admin who posts anonymously sails past the gate with full agent capability. A security keystone must be deny-by-default: anything not positively the owner is a bystander. **Fix:** when `OwnerID != 0`, treat `msg.From == nil` as a bystander, e.g.

   ```go
   if t.cfg.OwnerID != 0 {
       if msg.From == nil || !t.isOwner(msg.From.ID) {
           // bystander: decline only if addressed, else ignore; return
       }
   } else { /* one-time unset warn */ }
   ```

   The Discord side already gets this right (non-owner → `return` after optional decline, deny-by-default), so this is just bringing Telegram to the same posture. Please also add a test with `msg.From == nil` + `OwnerID` set asserting the agent is never reached — no current test covers the nil-sender path, which is why CI is green despite the hole.

### Nits (won't block)

1. `telegram.go:199` — the bystander decline has no per-user throttle. A non-owner can repeatedly @mention the bot in a shared group and make it emit `bystanderDecline` every time, which is a self-inflicted flood / Telegram rate-limit-ban vector. A short per-user cooldown or dedupe on the decline would harden it; the "only on direct address" rule bounds but doesn't prevent it.
2. `telegram.go:247-251` — the unterminated-header branch of `scrubCrosspostLiterals` rejoins as `cleaned[:s] + cleaned[s+nl:]`, leaving a doubled newline (`"line one\n\nline three"`). Cosmetic only; `TrimSpace` won't collapse the internal blank line.
3. `discord.go:~123` (pre-existing, in scope because this is the security pass on this handler) — `m.Author.ID` is dereferenced with no nil check (the existing self-message guard does it too); Discord system/auto-mod messages can have `m.Author == nil`. It's contained by the `recover()` defer, but a `if m.Author == nil { return }` at the top of the handler would be cleaner than relying on panic recovery on a security path.

### Things I like

- Pure `isOwner` classifier enforced at the platform boundary — no authority bit has to thread through the agent; clean and hard to misuse.
- Discord path is correctly deny-by-default for non-owners (DM or @mention → fixed decline, otherwise silent, always `return`).
- `sync.Once` for the unset-owner WARN — exactly one hint, no log spam.
- Backward compatible: `owner_id` unset = current behaviour, documented in config with the safety rationale.
- `scrubCrosspostLiterals` has genuinely good adversarial coverage (clean / unclosed header / stray close / unterminated) and is a sensible belt-and-braces over the directive parser.
- Scope discipline: the three deferred items have concrete, correct reasons (esp. not destabilising the `unified` channel without a migration).

### Verdict

**Re-requesting changes** for **item 1** — the `msg.From == nil` allow-by-default path is a real bypass of the single-owner safety keystone (anonymous-admin vector in shared groups), and it's untested. Nits won't block. Once the Telegram gate is deny-by-default with a nil-sender test, this is in good shape — the rest of the PR is solid.

srvsngh99:
## Review — PR #37 (round 2) — LGTM

**Checked at** head `f141df4` ("classify From==nil bystanders explicitly; scrub bare close tag"), reviewed in an isolated git worktree. `go build ./...`, `go vet ./...`, full `go test ./...`, `go test -race ./internal/chat/` all green; **CI fully green** (6 builds + Lint + Test). The round-1 blocking item is resolved.

### Round-1 item 1 (BLOCKING — `From==nil` bypassed the owner gate) — ✅ resolved

`telegram.go:184-213` — the gate is restructured to deny-by-default:

```go
if t.cfg.OwnerID == 0 {            // legacy: one-time warn, fall through
} else if msg.From == nil || !t.isOwner(msg.From.ID) {
    // bystander — incl. nil From (anonymous/sender_chat/linked-channel
    // auto-forward); decline only if addressed, else ignore
    return                          // unconditional — agent never reached
}
```

When `OwnerID != 0`, the *only* path that falls through to command/agent handling is `msg.From != nil && t.isOwner(msg.From.ID)`. `msg.From == nil` is now classified as a bystander **at the boundary** and `return`s there, so the safety property no longer depends on an incidental downstream nil-deref + `recover()` (the commit message correctly diagnoses that this was the prior reason it "held"). The `return` is now unconditional for both the addressed and ignored bystander sub-cases. Verified by inspection — the anonymous-group-admin / channel-sender vector is closed, and Telegram now matches Discord's deny-by-default posture.

### Beyond what was asked (good)

- **Nit 2 escalated into a real fix:** I only flagged the cosmetic double-newline, but self-review found a genuine leak — a bracket-less `[/CROSSPOST` close fragment survived both the early-out guard and the full-tag `ReplaceAll`. `telegram.go:1060` widens the guard to `[/CROSSPOST` and adds a trailing `ReplaceAll("[/CROSSPOST", "")`, with a `keep[/CROSSPOST` regression case. That's a correct catch and fix.
- **Discord `isReplyToMe` parity:** `discord.go:144-148` now treats a reply to the bot's own message as addressing it (with a defensive `m.ReferencedMessage.Author != nil` guard), matching Telegram.
- Bystander-decline-bypasses-bot-turn-accounting is now documented (single fixed reply, never reaches the agent → can't drive a loop).

### Non-blocking follow-ups (do not patch into this approved PR)

1. **The processUpdate-level regression test I asked for in round 1 was not added.** Coverage is the `isOwner` unit test + scrub cases; there is still no test driving `processUpdate` with `msg.From == nil` + `OwnerID` set asserting the agent is never reached. I'm not blocking on it: the hole is structurally closed and self-evident (a single unconditional `return` in the bystander branch), `isOwner` is unit-tested, and there's no existing `processUpdate` harness so adding one is a larger lift than belongs in this security fix. But it's genuinely owed — a future refactor of that gate has no test guarding the boundary. Worth a focused follow-up that introduces a `processUpdate` seam.
2. Cosmetic double-newline in the unterminated-header scrub branch (`"line one\n\nline three"`) remains — harmless.
3. `discord.go` still dereferences `m.Author.ID` without an explicit nil guard (pre-existing, `recover()`-contained). A `if m.Author == nil { return }` at the handler top would be cleaner than relying on panic recovery on a security path.

### Verdict

**LGTM — approving.** The blocking `From==nil` bypass is fixed correctly and the boundary is now self-contained and deny-by-default; CI and local build/vet/test/-race are green; a related real scrub leak was caught and fixed too. Follow-up 1 (the missing boundary regression test) is owed but non-blocking and explicitly should not re-trigger this PR. Not merging (separate explicit step).

PR #23, opened 2026-05-01
Add dedicated MCP tab to TUI

srvsngh99:
## Summary
- Extract MCP server management from the Skills tab into a dedicated **MCP** tab (tab position 4)
- MCP tab shows server connection status, tool counts, and supports enable/disable toggling with a detail panel
- Skills tab now focuses purely on skill registry — all MCP-specific code (`isMCP`, `mcpReg`) removed
- Tab bar updated to 6 tabs: Dashboard, Chat, Skills, **MCP**, Logs, Help (keys 1-6)

## Test plan
- [ ] `go build ./...` compiles cleanly
- [ ] Run `krill tui` — verify 6 tabs visible in the tab bar
- [ ] Press `4` to jump to MCP tab — shows server list (or empty state if none configured)
- [ ] Press `3` to jump to Skills tab — no MCP servers listed
- [ ] Tab/Shift+Tab wraps through all 6 tabs correctly
- [ ] Configure an MCP server in `config/default.yaml` and verify it appears in the MCP tab with connection status
- [ ] Toggle enable/disable on an MCP server with Enter/Space

PR #38, opened 2026-05-16
Provider honesty + secret-match hardening; v0.1.3

srvsngh99:
Final PR of the v0.1.3 cycle — the release PR. Scoped to bounded, low-risk hardening + the version bump; the large introspection suite is explicitly deferred (see below) rather than rushed.

## #23 — honest Codex catalog
Removed the fabricated Codex model list (`gpt-5.5`, `gpt-5.4`, `gpt-5.4-mini`, `gpt-5.3-codex`, `gpt-5.2`). Those IDs don't exist, so selecting one produced a confusing failure ~90s later. Codex now advertises only `auto`, delegating model choice to the Codex CLI (`cli_provider` already omits `--model` when the model is `auto`). Pruned the matching `modelMap` entries so `/use gpt-5.5` no longer resolves to codex.

## secretPathPatterns hardening (PR #32 nit)
The secret check was a raw substring over the whole path, so `notconfig.yaml.txt` was wrongly refused (it contains `config.yaml`). New `secretMatch` anchors each pattern to a real filename boundary: extension patterns match a true suffix, `secrets.` matches a basename prefix, and name tokens match the whole basename / a `.bak`-style variant / a full path component — never a mid-name substring. Real secrets (`config.yaml`, `.env`, `*.pem/.key`, `private_key`, `service_account`, `credentials/`) are still refused.

## Release
`core.Version` 0.1.2 → **0.1.3**, README badge, and a `CHANGELOG [0.1.3]` section covering PRs A, A2, B, and C.

## Deferred (documented, not dropped)
The big introspection items — `self:explain-decision` (E4.3), `self:dashboard` (#33), replay-from-log (#35), active-listening (#31) — are each a substantial new subsystem. Cramming four of them into the release PR in a multi-session shared repo would trade quality for speed. They move to a v0.1.4 introspection cycle. Also deferred with reasons: per-space context scoping (plan Risk #2), Telegram reply-to (E3.1), and a real #9 backoff (the live loop uses the library's `GetUpdatesChan`, not a tweakable retry). #7 was already largely resolved by the PR #27 `runCLI` rewrite.

## Test
`secretMatch` (refuse set + the `notconfig.yaml.txt` regression + allow set). `go build`/`vet`/`test ./...` all green; binary reports `minikrill version 0.1.3`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,31 @@
 All notable changes to Mini Krill will be documented in this file.
 Format based on [Keep a Changelog](https://keepachangelog.com/).
 
+## [0.1.3] - 2026-05-16
+
+Closed-loop learning, cross-session memory, single-owner safety, and an honest provider catalog. The v0.1.3 cycle (PRs A, A2, B, C).
+
+### Closed-loop learning (PR A)
+- Post-turn THANKED/FIXED callback. The affinity learner was one-directional — an auto-run produced no verdict, so a calibrated cluster's score could only ever rise and a cluster that started producing bad output stayed auto-run forever. The user's next message after an un-gated turn is now the verdict: a correction credits FIXED (drift back toward planning), anything else credits THANKED. Restores symmetric drift; closes the freeze flagged in the PR #33 review.
+- #16: `looksLikeCorrection` caught `"don't"` but not `doesn't`/`didn't`/`isn't`/`wasn't`/`won't`/`can't`/… so real corrections went uncredited. Extended the contraction list (word-bounded) plus curly-apostrophe normalisation.
+- `affinity:cluster:` is now a reserved memory namespace: `FileMemory.Store` refuses writes there unless they come from the affinity store, so `/remember` can't corrupt a learned record.
+
+### Cross-session memory & user-state (PR A2)
+- Episodic memory (#29 / D6): on a >30 min activity gap the prior session is summarised into one episode; the most recent episode <7 days old is injected into chat context as an explicitly point-in-time, read-only line so the agent resumes instead of starting cold.
+- User-state (#37): focus area, last task cluster, light mood, last-seen — persisted and injected read-only as point-in-time context.
+
+### Single-owner safety (PR B)
+- New `telegram.owner_id` / `discord.owner_id`. When set, only the owner drives the bot; a bystander in a shared group/server is replied to with a fixed decline only if they directly address the bot, and their message never reaches the agent — no tasks, memory writes, or destructive actions from a stranger. The safety counterpart to "no approval prompts by default". Unset = legacy behaviour with a one-time hint.
+- #22: residual `[CROSSPOST:..]` literals from a malformed directive are scrubbed so the raw token never leaks into chat.
+
+### Provider honesty & hardening (PR C)
+- #23: removed the fabricated Codex model catalog (`gpt-5.5`, `gpt-5.4`, …). Those IDs don't exist and picking one failed ~90s later. Codex now advertises only `auto`, delegating model choice to the Codex CLI.
+- `secretPathPatterns` (self:read-code / read-logs) anchored to real filename boundaries instead of a raw substring, so a benign `notconfig.yaml.txt` is no longer refused while real secrets still are.
+
+### Notes
+- #7 (Codex stderr/timeout surfacing) was largely resolved by the PR #27 `runCLI` rewrite (separate stderr, distinct ctx-cancel surfacing, idle monitor).
+- Deferred to a future introspection cycle: `self:explain-decision`, `self:dashboard`, replay-from-log, active-listening primitive, per-space context scoping, Telegram reply-to capture, and a real Telegram backoff (the live polling loop uses the library's `GetUpdatesChan`).
+
 ## [0.1.2] - 2026-05-16
 
 No more nagging. The agent stops asking for manual permission on routine work.
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@
 
 Built by [Sourav Singh](https://souravailabs.ai/about/) / [Sourav AI Labs](https://souravailabs.ai)
 
-[![Version](https://img.shields.io/badge/version-0.1.2-blue.svg)](https://github.com/srvsngh99/mini-krill/releases)
+[![Version](https://img.shields.io/badge/version-0.1.3-blue.svg)](https://github.com/srvsngh99/mini-krill/releases)
 [![Go](https://img.shields.io/badge/Go-1.24+-00ADD8.svg?logo=go)](https://go.dev/)
 [![License](https://img.shields.io/badge/license-MIT-green.svg)](LICENSE)
 [![Platforms](https://img.shields.io/badge/platforms-Linux%20%7C%20macOS%20%7C%20Windows-lightgrey.svg)]()
--- a/internal/chat/telegram.go
+++ b/internal/chat/telegram.go
@@ -523,7 +523,7 @@ func (t *TelegramBot) handleCommand(chatID int64, msg *tgbotapi.Message) {
 			"/status  - Check my vital signs",
 			"/model   - Show active provider and model",
 			"/models  - List all providers and models",
-			"/switch  - Switch provider (e.g. /switch local, /switch codex gpt-5.5)",
+			"/switch  - Switch provider (e.g. /switch local, /switch codex)",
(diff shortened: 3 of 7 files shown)


srvsngh99:
## Review — PR #38 (round 1)

**Checked at** head `0889707`, reviewed in an isolated git worktree. `go build ./...`, `go vet ./...`, full `go test ./...`, `go test -race ./internal/plugin/` all green; CI green. The version bump (0.1.2 → 0.1.3 in `core/types.go` + README badge + CHANGELOG `[0.1.3]` covering A/A2/B/C) is correct and matches the patch-only, bump-in-the-final-PR cadence. But the `secretMatch` hardening introduces a real secret-protection **regression**, and #23 is only half-done.

### Substantive issues

1. **`secretMatch` no longer refuses `.env.local` / `.env.production` / `.env.<env>` — a secret-protection regression (blocking).** `self_eyes.go:187-190`: extension-like patterns (those starting with `.`) only match via `strings.HasSuffix(base, pat)`. I verified empirically against the actual code:

   ```
   /repo/.env.local           refused=false
   /repo/.env.production      refused=false
   /repo/app/.env.development refused=false
   /repo/server.pem.bak       refused=false
   /repo/.env                 refused=true
   /repo/prod.env             refused=true
   ```

   The old raw-substring check (`strings.Contains(abs, ".env")`) **did** refuse `.env.local` et al. The dotenv multi-environment convention (`.env.local`, `.env.production`, `.env.development`, `.env.test`) is the *most common* real secret-file layout, and `self:read-code` / `read-logs` will now happily read them. Hardening the matcher to fix a false-positive must not open a false-negative on the highest-value case. **Fix:** for extension-like patterns, also anchor on the basename as a dot-segment, e.g. match if `base == pat || strings.HasPrefix(base, pat+".") || strings.HasSuffix(base, pat)` — the `HasPrefix(base, ".env.")` arm catches `.env.local`/`.env.production`, the suffix arm keeps `prod.env`, and neither reintroduces the `notconfig.yaml.txt` false positive (that's a `config.yaml` whole-name token, unaffected). Add `.env.local`, `.env.production`, and a `*.pem.bak` to the refuse set in `secret_match_test.go` — the current test omits exactly the cases that regressed, which is why CI is green.

2. **#23 is incomplete — user-facing help still advertises the fabricated `gpt-5.5`.** The PR removes `gpt-5.5`/`gpt-5.4`/… from `ListProviders` and `modelMap`, but `telegram.go:526` (`/switch  - Switch provider (e.g. /switch local, /switch codex gpt-5.5)`) and `telegram.go:697` (`Example: /switch codex gpt-5.5`) still tell the user to type a model ID that no longer resolves. That reproduces the exact ~90-second confusing failure #23 exists to eliminate — arguably worse, because the bot's own help text is now the thing recommending the dead ID. **Fix:** change both strings to `/switch codex` (or `/switch codex auto`) so the documented contract matches the honest catalog.

### Nits (won't block)

1. Coverage asymmetry between directory-name secrets: `credentials` matches a full path component (`/credentials/…`) but `secrets.` only matches a basename prefix, so a `secrets/` *directory* (`/repo/secrets/db.txt`) is not refused. The old substring check didn't catch it either, so this isn't a regression — but while you're in here, treating `secrets` as a path-component token too would close it.
2. `application.config.yaml` is also no longer refused (old substring caught it via `config.yaml`). Lower-value than item 1 and naming-convention-dependent; the item 1 fix pattern (dot-segment anchoring) could be applied to whole-name tokens too if you want symmetry, but I won't block on it.

### Things I like

- `secretMatch` returning the matched pattern (vs a bare bool) makes the refusal message actionable, and the basename/extension/prefix/component split is the right structure — it just needs the `.env.<x>` arm.
- The honest-catalog change is the right call: `Models: ["auto"]` with a comment explaining the delegation to the Codex CLI, and `modelMap` pruned so `/use gpt-5.5` can't mis-resolve. Just finish it in the help text.
- CHANGELOG `[0.1.3]` is thorough and accurately attributes A/A2/B/C, and correctly records #7 as resolved-by-#27 and the deferred introspection items with reasons.
- Version bump lands only in this final cycle PR — consistent with the established cadence.

### Verdict

**Re-requesting changes** for **item 1** (the `.env.local`/`.env.*` secret-read regression — blocking; empirically confirmed) and **item 2** (#23 left the fabricated `gpt-5.5` in the user-facing `/switch` help, reproducing the failure it claims to fix). Nits won't block. Everything else — the version/CHANGELOG/README release mechanics and the catalog data change — is correct and shouldn't regress in round 2.

srvsngh99:
## Fixes — PR #38 (round 1 response)

Pushed `c9fee47` (was `0889707`). `go build`/`vet`/`test ./...` green; `TestSecretMatch` passes with the regressed cases now in the refuse set.

**Item 1 (blocking — `.env.*` / `*.pem.bak` secret-read regression): fixed.** `self_eyes.go` extension-like arm previously only matched a true suffix. Added an interior dot-segment match: `strings.HasSuffix(base, pat) || strings.Contains(base, pat+".")`. Now refused again: `.env.local`, `.env.production`, `app/.env.development`, `server.pem.bak`. Still allowed (no regression): `notconfig.yaml.txt` (a `config.yaml` whole-name token, untouched), `envparser.go`, `keyboard.go` (no dot-anchored segment — `.env.`/`.key.` not present). Added all four regressed paths to the `refuse` set in `secret_match_test.go` so CI covers exactly the cases that were silently green before.

**Item 2 (#23 incomplete — `gpt-5.5` still in `/switch` help): fixed.** `telegram.go:526` and `telegram.go:697` now read `/switch codex` instead of `/switch codex gpt-5.5`. The bot's own help text no longer recommends the dead ID.

**Nits:** not patched (won't block, per convention — left as follow-ups so an approved PR isn't re-triggered).

Re-requesting review.

srvsngh99:
## Review — PR #38 (round 2) — LGTM

**Checked at** head `c9fee47` ("refuse .env.* / *.pem.bak again; finish #23 in /switch help"), reviewed in an isolated git worktree. `go build ./...`, `go vet ./...`, full `go test ./...`, `go test -race ./internal/plugin/` all green; **CI fully green** (6 builds + Lint + Test). Both round-1 items resolved, empirically confirmed.

### Round-1 item 1 (BLOCKING — `.env.local`/`.env.*`/`*.pem.bak` secret-read regression) — ✅ resolved

`self_eyes.go:62` — the extension-like arm is now `strings.HasSuffix(base, pat) || strings.Contains(base, pat+".")`. The interior dot-segment match closes the regression without reopening the original false positive. I re-probed `secretMatch` directly against the actual code:

```
/r/.env.local           refused=true  pat=".env"
/r/.env.production      refused=true  pat=".env"
/r/app/.env.development refused=true  pat=".env"
/r/server.pem.bak       refused=true  pat=".pem"
/r/.env                 refused=true  pat=".env"
/r/prod.env             refused=true  pat=".env"
/r/envparser.go         refused=false        (still readable ✓)
/r/keyboard.go          refused=false        (still readable ✓)
/r/notconfig.yaml.txt   refused=false        (still readable ✓)
/r/configurator.md      refused=false        (still readable ✓)
```

The dotenv multi-environment convention and `.pem` backups are refused again; `envparser.go`/`keyboard.go` (no dot-anchored segment) and the original `notconfig.yaml.txt` (a `config.yaml` whole-name token, untouched by this arm) stay readable. `secret_match_test.go` now carries `.env.local`, `.env.production`, `app/.env.development`, and `server.pem.bak` in the refuse set, so CI covers exactly the cases that regressed. `TestSecretMatch` passes.

### Round-1 item 2 (#23 incomplete — fabricated `gpt-5.5` in `/switch` help) — ✅ resolved

`telegram.go:526` and `telegram.go:697` now read `/switch codex` (no model ID). A repo-wide `grep gpt-5` over non-test `.go` files returns nothing except the two intentional `manager.go` #23 explanatory comments — no advertised fabricated ID remains anywhere user-facing.

### Verdict

**LGTM — approving.** The secret-read regression is closed (verified by direct re-probe, not just the test), the false-positive guard the hardening originally existed for is intact, and #23 is now consistent end-to-end including the help text. CI and local build/vet/test/-race green. Version/CHANGELOG/README release mechanics were validated in round 1 and are unchanged. Not merging (separate explicit step).

srvsngh99:
Thanks — round-2 LGTM/approval received at head `c9fee47`. Both round-1 blocking items confirmed resolved; review loop closed, polling stopped. Not merging (separate explicit step).

PR #39, opened 2026-05-16
PR37 follow-ups: owner-gate seam + scrub blank line + discord nil-author guard

srvsngh99:
Addresses the three **non-blocking follow-ups** from the #37 review (deliberately a separate PR — not patched into the approved one, per the review's own instruction).

## 1. (owed) Owner-gate test seam — `internal/chat/telegram.go`
The round-1 review's owed item: there was no test driving the owner boundary with `msg.From == nil` + `owner_id` set. Extracted the decision into a pure, table-tested classifier:

```go
decideOwnerGate(ownerSet, isOwner, addressed bool) ownerGate // → proceed | ignore | decline
```

`processUpdate` now `switch`es on it. Deny-by-default: anything that isn't a confirmed owner under an active gate is ignored/declined, never proceeded. `TestDecideOwnerGate` asserts the owed invariant directly — *owner-gating on + non-owner (the `From==nil` regime) never resolves to `gateProceed`*, addressed or not — so a future refactor of that gate has a guard instead of relying on an unconditional `return` + incidental downstream `recover()`.

## 2. Scrub blank-line — `internal/chat/telegram.go`
`scrubCrosspostLiterals`'s unterminated-header branch cut the broken segment but kept its trailing newline, leaving `"line one\n\nline three"`. Now consumes the newline too; the existing scrub test asserts no `\n\n` residue.

## 3. Discord nil-author guard — `internal/chat/discord.go`
Explicit `if m.Author == nil { return }` at the handler top instead of relying on panic recovery on a security path (the owner gate dereferences `m.Author`).

## Scope / test
No version bump — the v0.1.3 bump belongs to the cycle's final PR. `go build`/`vet`/`test ./...` and `go test -race ./internal/chat/` all green.

### DIFF
--- a/internal/chat/discord.go
+++ b/internal/chat/discord.go
@@ -118,6 +118,13 @@ func (d *DiscordBot) onMessageCreate(ctx context.Context) func(s *discordgo.Sess
 			}
 		}()
 
+		// Defensive: system/webhook edge cases can carry a nil Author. Drop
+		// explicitly rather than relying on the panic recovery above — this
+		// is a security path (the owner gate below dereferences Author).
+		if m.Author == nil {
+			return
+		}
+
 		// Never reply to ourselves.
 		if m.Author.ID == s.State.User.ID {
 			return
--- a/internal/chat/owner_gating_test.go
+++ b/internal/chat/owner_gating_test.go
@@ -7,6 +7,39 @@ import (
 	"github.com/srvsngh99/mini-krill/internal/config"
 )
 
+// TestDecideOwnerGate is the boundary regression owed from PR #37 review
+// (follow-up 1): with owner-gating on, anything that is not a confirmed owner
+// — crucially the msg.From == nil case, which presents as isOwner=false —
+// must never resolve to gateProceed, so the agent is never reached.
+func TestDecideOwnerGate(t *testing.T) {
+	cases := []struct {
+		name                       string
+		ownerSet, isOwner, address bool
+		want                       ownerGate
+	}{
+		{"gating off → always proceed", false, false, true, gateProceed},
+		{"gating off, unaddressed → proceed", false, false, false, gateProceed},
+		{"owner proceeds", true, true, false, gateProceed},
+		{"owner proceeds even if addressed", true, true, true, gateProceed},
+		{"bystander addressing → decline", true, false, true, gateDecline},
+		{"bystander silent → ignore", true, false, false, gateIgnore},
+	}
+	for _, c := range cases {
+		if got := decideOwnerGate(c.ownerSet, c.isOwner, c.address); got != c.want {
+			t.Errorf("%s: decideOwnerGate(%v,%v,%v)=%d want %d",
+				c.name, c.ownerSet, c.isOwner, c.address, got, c.want)
+		}
+	}
+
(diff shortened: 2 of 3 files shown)


srvsngh99:
## Review — PR #39 (round 1) — LGTM

**Checked at** head `5ebc8af`, reviewed in an isolated git worktree. `go build ./...`, `go vet ./...`, full `go test ./...`, and `go test -race ./internal/chat/` all green. Three non-blocking #37 follow-ups, correctly split out of the approved PR rather than patched into it.

### 1. Owner-gate test seam (`telegram.go`) — ✅

`decideOwnerGate(ownerSet, isOwner, addressed)` is pure (no tgbotapi/I/O) and deny-by-default: `!ownerSet || isOwner → proceed`, else `addressed → decline`, else `ignore`. I checked the `processUpdate` switch against the old nested branch for every case — legacy (`OwnerID==0` → warn + proceed), owner (proceed), bystander-addressed (decline + return), bystander-silent (ignore + return) — behaviorally identical. The owed invariant is asserted directly: `telegram.go:54` loops `decideOwnerGate(true, false, addressed)` over both `addressed` values and fails on any `gateProceed`, so the `msg.From == nil` regime (which presents as `isOwner=false`) can never reach the agent. The safety property is now the gate's, not an incidental downstream `recover()`.

### 2. Scrub blank-line (`telegram.go:1117`) — ✅

`nl := strings.IndexByte(cleaned[s:], '\n')` is relative to `cleaned[s:]`, so the newline is at absolute `s+nl`; `cleaned[:s] + cleaned[s+nl+1:]` consumes it and joins cleanly with no `\n\n`. `s+nl+1 ≤ len(cleaned)` always (it's a found index, only entered when `nl != -1`), so no slice panic. `TestScrubCrosspostLiterals` now asserts no blank-line residue.

### 3. Discord nil-author guard (`discord.go:121`) — ✅

`if m.Author == nil { return }` sits immediately after the deferred recover and before the first `m.Author.ID` deref / the owner gate. A nil-author system/webhook message is now dropped explicitly on the security path instead of relying on panic recovery. Correct placement.

### Nits (won't block)

1. `telegram.go` — the `var uid, uname int64/string; if msg.From != nil { uid, uname = ... }` block is duplicated verbatim in `gateDecline` and `gateIgnore`. Could hoist it once before the `switch` (the values are only read in those two arms). Cosmetic.
2. `isOwner`/`addressed` are now computed for *every* message, including the owner's and in legacy mode, where the old code computed them lazily inside the bystander branch. Both are pure reads (`isMentioned` + a guarded reply-chain walk), so it's a harmless micro-expansion, not a regression — noting only for completeness.

### Things I like

- The pure-classifier extraction is exactly the right shape for the owed test: the boundary is now guarded by an assertion instead of an unconditional `return` + downstream `recover()`.
- Deny-by-default is structural (the `switch` has no fall-through to handling except `gateProceed`), not incidental.
- Correctly scoped: no version bump (the v0.1.3 bump stays in the cycle's release PR), and split from the approved #37 per that review's own instruction.
- Tests target exactly the three changed behaviors and nothing more.

### Verdict

**LGTM — approving.** All three follow-ups are correct, minimal, and tested; build/vet/test/-race green. Nits are cosmetic and won't block. Not merging (separate explicit step).

PR #40, opened 2026-05-16
Fix silent identity loss on runtime rename

srvsngh99:
## Problem

The user-chosen `agent_name` and `personality` (set at `minikrill init`) could be **silently erased** by any runtime config save — observed in the wild as a config that drifted to `agent_name: aaaa…` with the personality reset to `krill`.

Two independent causes:

1. **`MarshalYAML` tagged `agent_name`/`personality` `omitempty`** — a save where the in-memory field was empty dropped the line from `config.yaml` entirely. The next `Load` then defaulted personality to `"krill"` (`fillDefaults`), a silent identity downgrade with no error logged.
2. **`persistAgentConfig` did a whole-block `cfg.Agent = agent` replace** — a `/name` rename carries an `AgentConfig` whose `Personality` is unset, so persisting a rename clobbered the on-disk personality.

`init` itself was never at fault; it writes both fields correctly. The loss happened post-init via the runtime rename path.

## Fix

- Drop `omitempty` on the two identity fields so a set identity always round-trips and can never be silently dropped.
- `persistAgentConfig` now preserves the loaded `personality`/`agent_name` when the incoming config leaves them blank, instead of whole-block clobbering.

## Tests

Adds `identity_persist_test.go`:
- rename with blank personality must preserve the on-disk personality (and still apply the new name)
- `agent_name`/`personality` round-trip through `Save`/`Load`

`go build ./...`, `go test ./internal/agent ./internal/config`, `go vet` all green.

### DIFF
--- a/internal/agent/identity.go
+++ b/internal/agent/identity.go
@@ -297,6 +297,17 @@ func persistAgentConfig(agent config.AgentConfig) error {
 	if err != nil {
 		return err
 	}
+	// Whole-block replace would clobber identity fields the caller didn't
+	// intend to touch: a /name rename carries an in-memory AgentConfig
+	// whose Personality may be empty, which previously erased the
+	// init-chosen personality on disk. Preserve the loaded value for any
+	// identity field the incoming config leaves blank.
+	if agent.Personality == "" {
+		agent.Personality = cfg.Agent.Personality
+	}
+	if agent.AgentName == "" {
+		agent.AgentName = cfg.Agent.AgentName
+	}
 	cfg.Agent = agent
 	return config.Save(cfg)
 }
--- a/internal/agent/identity.go
+++ b/internal/agent/identity.go
@@ -302,6 +302,16 @@ func persistAgentConfig(agent config.AgentConfig) error {
 	// whose Personality may be empty, which previously erased the
 	// init-chosen personality on disk. Preserve the loaded value for any
 	// identity field the incoming config leaves blank.
+	//
+	// Design: empty means "leave whatever is on disk", NOT "clear it".
+	// Clearing identity is not a supported operation anywhere in the code
+	// (RenameAgent rejects empty input; SetPersonality always sets a
+	// concrete name), so there is no caller that legitimately wants an
+	// empty value persisted. This is the same intent as the matching
+	// non-omitempty tags in config.MarshalYAML: both refuse to let an
+	// accidental blank silently destroy user-chosen identity. If a
+	// deliberate clear is ever needed, it must be a distinct explicit
+	// path, not an empty field threaded through this helper.
 	if agent.Personality == "" {
 		agent.Personality = cfg.Agent.Personality
 	}
--- a/internal/agent/identity_persist_test.go
+++ b/internal/agent/identity_persist_test.go
@@ -1,0 +1,60 @@
+package agent
+
+import (
+	"testing"
+
+	"github.com/srvsngh99/mini-krill/internal/config"
+)
+
+// Regression: a /name rename carries an in-memory AgentConfig whose
+// Personality is empty. persistAgentConfig must NOT let that erase the
+// init-chosen personality on disk (it previously did, after which Load
+// silently defaulted personality to "krill" — a silent identity downgrade).
+func TestPersistAgentConfig_PreservesPersonalityOnRename(t *testing.T) {
+	t.Setenv("KRILL_DATA_DIR", t.TempDir())
+
+	seed := config.DefaultConfig()
+	seed.Agent.Name = "test-krill"
+	seed.Agent.AgentName = "Jarvis"
+	seed.Agent.Personality = "jarvis"
+	if err := config.Save(seed); err != nil {
+		t.Fatalf("seed Save: %v", err)
+	}
+
+	// Simulate RenameAgent's call: only AgentName set, Personality blank.
+	if err := persistAgentConfig(config.AgentConfig{AgentName: "Friday"}); err != nil {
+		t.Fatalf("persistAgentConfig: %v", err)
+	}
+
+	got, err := config.Load()
+	if err != nil {
+		t.Fatalf("Load: %v", err)
+	}
+	if got.Agent.Personality != "jarvis" {
+		t.Errorf("personality = %q, want %q (must survive a rename)", got.Agent.Personality, "jarvis")
+	}
+	if got.Agent.AgentName != "Friday" {
+		t.Errorf("agent_name = %q, want %q (rename must still apply)", got.Agent.AgentName, "Friday")
+	}
+}
+
+// Regression: agent_name/personality must be written even when set, never
+// dropped by omitempty — a round-trip through Save/Load preserves them.
+func TestMarshalYAML_IdentityFieldsNotOmitted(t *testing.T) {
+	t.Setenv("KRILL_DATA_DIR", t.TempDir())
+
+	c := config.DefaultConfig()
+	c.Agent.AgentName = "Jarvis"
+	c.Agent.Personality = "jarvis"
+	if err := config.Save(c); err != nil {
+		t.Fatalf("Save: %v", err)
+	}
+	got, err := config.Load()
+	if err != nil {
+		t.Fatalf("Load: %v", err)
+	}
+	if got.Agent.AgentName != "Jarvis" || got.Agent.Personality != "jarvis" {
+		t.Errorf("identity not round-tripped: agent_name=%q personality=%q",
+			got.Agent.AgentName, got.Agent.Personality)
+	}
+}
--- a/internal/agent/identity_persist_test.go
+++ b/internal/agent/identity_persist_test.go
@@ -1,6 +1,9 @@
 package agent
 
 import (
+	"os"
+	"path/filepath"
+	"strings"
 	"testing"
 
 	"github.com/srvsngh99/mini-krill/internal/config"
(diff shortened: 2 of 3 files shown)


srvsngh99:
## Review - PR #40 (round 1)

Ran on head SHA `069b611` (branch `fix/identity-persistence`): `gofmt -l` (clean), `go build ./...` (pass), `go vet ./internal/agent/ ./internal/config/` (pass), `go test ./internal/agent/ ./internal/config/` (pass). Also reverted the fix locally and re-ran the new tests to confirm they catch the bug; audited other `config.Save`/`persistAgentConfig` call sites and the YAML round-trip.

The core fix is correct and well-targeted: dropping `omitempty` on `agent_name`/`personality` plus the blank-preservation guard in `persistAgentConfig` does stop the silent identity downgrade, and `TestPersistAgentConfig_PreservesPersonalityOnRename` genuinely fails without the fix. There are issues to address before merge.

### Substantive issues

1. **Em dashes added to the codebase and commit message (hyphen-only required).** `internal/config/config.go:95`, `internal/agent/identity_persist_test.go:12`, `internal/agent/identity_persist_test.go:42`, and the commit message body all introduce U+2014 em dashes (e.g. the string `"krill" <U+2014> a silent identity downgrade`) on new/added lines. This repo's standard is ASCII hyphens only; the PR adds new non-ASCII punctuation rather than only inheriting pre-existing instances. Fix: replace each em dash on an added line with ` - ` (space-hyphen-space), and amend the commit message to use a hyphen as well, then force-push.

2. **`TestMarshalYAML_IdentityFieldsNotOmitted` does not actually exercise the `omitempty` removal.** `internal/agent/identity_persist_test.go:46-60`: the test sets `AgentName`/`Personality` to non-empty values before `Save`, so `omitempty` would be a no-op anyway - I verified this by restoring `omitempty` and the test still passed (only the rename test failed). The regression this test claims to guard ("never dropped by omitempty") is not covered: the dangerous case is an *empty* identity field being silently dropped on write and then defaulting on load. Fix: add an assertion that exercises an empty-field save - e.g. marshal an `AgentConfig` with `Personality: ""` directly and assert the `personality:` key is still present in the YAML bytes (or save a config whose in-memory `Personality` is "" and assert it round-trips to "" rather than disappearing), so the test fails if `omitempty` is reintroduced.

3. **The blank-preservation guard cannot represent a deliberate clear, and no caller path needs that today, but the asymmetry with the `omitempty` rationale should be reconciled.** `internal/agent/identity.go:305-310`: the empty-string check means a future/legitimate "reset agent_name to default" path silently becomes a no-op (the old on-disk value is restored). The PR's own `config.go` comment argues empties should be *written* so loss is "visible", yet `persistAgentConfig` does the opposite (hides the empty by back-filling). Both behaviors are defensible, but the inconsistency is undocumented. There is currently no call site that intentionally clears these fields (`RenameAgent` always sets a non-empty validated name; `handlePersonalityCommand` always sets a non-empty `rest`), so this is not a live bug - but please add a one-line comment at `identity.go:305` stating that an empty incoming identity field is treated as "no change, preserve disk" by design, and that callers must not rely on `persistAgentConfig` to clear identity.

### Nits

- `internal/agent/identity.go:295` `persistAgentConfig` takes `agent config.AgentConfig` by value and mutates the parameter; reusing the inbound name as a working copy is fine in Go but a short `// mutate local copy` would aid the next reader given the new back-fill logic.
- `RecoveryTurns` keeps `omitempty` while the two identity fields drop it. This is intentional and harmless (RecoveryTurns has a numeric default applied in `fillDefaults`, identity does not), but a half-sentence in the `MarshalYAML` comment noting why RecoveryTurns is treated differently would prevent a future "make it consistent" regression.

### Verified good

- `gofmt`, `go build ./...`, `go vet`, and targeted `go test ./internal/agent/ ./internal/config/` all pass at `069b611`.
- `TestPersistAgentConfig_PreservesPersonalityOnRename` is a real regression test: it fails (`personality = "krill", want "jarvis"`) when the fix is reverted.
- Other `config.Save` call sites in `internal/plugin/self.go` operate on a fully-loaded `sc.Config` that retains identity fields, so they do not reintroduce the clobber; not in scope and not regressed.
- Branch is `fix/identity-persistence` (not main/master); no secrets added; no `--no-verify`; no AI attribution / Co-Authored-By / tool names in the diff or commit message.
- The `omitempty` removal does not break existing on-disk configs: `Load` -> `fillDefaults` still defaults an empty `personality` to `krill`, and an extra always-written key is backward compatible for YAML readers.

### Verdict

CHANGES_REQUESTED - the fix is sound, but the em dashes violate the hyphen-only bar and the second regression test does not actually guard the `omitempty` behavior it claims to.

srvsngh99:
Addressed all 3 round 1 items in d395d67:

1. Em dashes removed. The flagged occurrences on this PR's added lines (config.go comment, the new test file) are now ASCII hyphens / reworded. Pre-existing em dashes elsewhere in identity.go are untouched as they are not part of this diff. The final squash-merge commit message will be authored clean.
2. TestMarshalYAML_IdentityFieldsNotOmitted rewritten: it now saves a config with empty agent_name/personality and asserts both keys are still emitted to config.yaml. Verified it FAILS with `,omitempty` restored and PASSES without it, so it genuinely pins the regression. Since the brain block also emits `personality:`, the check requires >=2 occurrences to catch the agent-side drop specifically.
3. persistAgentConfig now documents the blank-preservation contract: an empty incoming field means "keep the on-disk value", never "clear it"; clearing identity is not a supported operation (RenameAgent rejects empty, SetPersonality always sets a concrete name), which is the same intent as the non-omitempty marshal tags.

gofmt clean, go build ./..., go vet, and targeted go test ./internal/agent ./internal/config all pass.

srvsngh99:
## Review - PR #40 (round 2)

Head SHA: `d395d67` (branch `fix/identity-persistence`). Ran: `gofmt -l` on the 3 changed files (clean), `go build ./...` (clean), `go vet ./internal/agent/ ./internal/config/` (clean), `go test ./internal/agent/ ./internal/config/` (PASS). Added-line dash scan `git diff origin/main -- internal/ | grep -nP '[\x{2014}\x{2013}]'` empty. Commit-message attribution/secret scans empty. Branch is not main/master.

I independently re-derived the omitempty guard: copied config.go, restored `,omitempty` on both MarshalYAML fields, and ran the new tests. `TestMarshalYAML_IdentityFieldsNotOmitted` FAILED as designed (`agent_name: key missing`, `found 1 personality: lines`); restored the file and it PASSES. Also dumped the generated config.yaml: with omitempty removed there are exactly two `personality:` lines (agent block + brain block, the latter never omitempty per config.go:126), so the `>=2` assertion is keyed to a deterministic, Save-generated schema rather than free-form text.

### Substantive issues

None. All three round 1 items are genuinely fixed:

1. **Dashes (FIXED).** No em/en dash on any added line of the diff (`internal/agent/identity.go`, `internal/agent/identity_persist_test.go`, `internal/config/config.go`); commit messages `d395d67` / `069b611` are ASCII-clean. The two pre-existing em-dashes at `internal/config/config.go:86` and `internal/agent/identity.go:90` are out of scope — `git blame` attributes line 86 to commit `1a054ca` and neither line is a `+` line in this PR's diff.

2. **omitempty guard test now real (FIXED).** `internal/agent/identity_persist_test.go:67` and `:74` assert the keys are emitted for an empty struct. Independently verified the test fails with `,omitempty` restored and passes without it (see header). The `strings.Count(text, "personality:") >= 2` check is robust here: `config.Save` always emits the brain block's non-omitempty `personality:` (exactly one), so the count is 2 only when the agent field is also written; `agent_name:` is unique to the agent block so `Contains` suffices. The check is coupled to the brain block always emitting `personality:`, but that coupling is explicitly documented in the test comment at `identity_persist_test.go:69-73`, so it is self-explaining rather than a hidden trap. `>=2` (not `==2`) is appropriately lenient.

3. **Blank-preservation design documented + asymmetry reconciled (FIXED).** `internal/agent/identity.go:306-314` states "empty means leave-on-disk, NOT clear", explains why no caller wants an empty persisted (RenameAgent rejects empty input — verified at `identity.go:95-97`; personality switch sets a concrete value — verified at `identity.go:452`), and explicitly ties the intent to the non-omitempty MarshalYAML tags. Both real callers (`identity.go:104`, `identity.go:453`) pass a fully-populated `a.cfg`, and `config.Load` -> `fillDefaults` (config.go:305-306) guarantees a non-empty on-disk personality, so the guard is sound defensive behaviour with no new regression.

### Nits

- `internal/agent/identity_persist_test.go:42` doc comment has an unbalanced paren: `... defaulted personality to "krill" (a silent identity downgrade).` — opens `(` without a closing `)`. Cosmetic only.
- The `personality:` count assertion is structurally coupled to the brain block always serialising `personality:`. Fine today and documented, but a future `omitempty` on `BrainConfig.Personality` would silently weaken this guard. A scoped check (e.g. asserting against only the `agent:` block, or anchoring on `agent_name:` adjacency) would be more durable. Optional.

### Verified good

- gofmt / build / vet / targeted tests all green at `d395d67`.
- Test file imports are stdlib + `internal/config` only; no new third-party dependency.
- No secrets, API keys, or AI/tool attribution in diff or commit messages.
- Round 1 omitempty fix independently reproduced by restore-and-rerun (fails with omitempty, passes without).
- Guard comment factual claims (RenameAgent empty rejection, personality-switch concrete value) verified against source.
- Pre-existing em-dashes correctly left untouched and confirmed not introduced by this PR.

Verdict: APPROVED — all three round 1 items genuinely fixed and independently verified; only cosmetic nits remain.

srvsngh99:
Approved after 2 automated review rounds. Round 1 flagged dashes, a non-exercising omitempty test, and an underdocumented preservation guard; all fixed in d395d67 and verified in round 2 (the test was confirmed to fail with omitempty restored). Merging.

PR #41, opened 2026-06-13
KrillLM default provider (gemma 12b) + chat-history reliability fixes (v0.1.4)

srvsngh99:
## What the user asked for

Verbatim, across the session:
- "We have to support KrillLM and use gemma12b model as primary for this so make that change and bring up minikrill"
- "I want krilllm as default inference for mini krill and ollama and other providers are secondary and use the Gemma 12b model"
- "check the last chat messages and identify the issue and fix them" / "Analyze all the chat history and confirm what still exists and what has been fixed"
- "are we also updating the release notes and readme and adding support of Krilllm"

## Summary

Makes **KrillLM** Mini Krill's default local inference (Ollama-backed, primary model **gemma 12b** / `gemma4:12b-mlx`), with Ollama and the CLI/API providers demoted to secondary. Along the way this ships a real AI-digest skill and a batch of reliability fixes drawn from a full audit of the live chat history (`conversations.jsonl` + both daemon logs).

Bumps to **v0.1.4** (also reconciles the stale `VERSION` file, which read 0.1.1 while `core.Version`/CHANGELOG were at 0.1.3).

## KrillLM default provider
- New `krilllm` provider (`internal/llm/krilllm.go`) — Ollama-backed, `Name()=krilllm`, primary model gemma 12b; empty/`auto` resolves to the primary.
- Default everywhere: `DefaultConfig`, `config/default.yaml`, first/recommended `init` option.
- Switching surfaces learn it: `ResolveTarget` aliases (`krilllm`/`krill`/`krill-lm`, `gemma12b` shorthand), `self:configure`, Telegram `/use`. Ollama auto-start/health gating moved to `config.IsLocalProvider`.

## Daemon reliability
- **Silent death of detached `dive`** — child output was piped through an in-process writer; the pipe collapsed (SIGPIPE) when the parent exited after `Release()`. Child now inherits the log fd directly.
- **`krill.pid` = -1** — pid was read after `Release()`. Captured before now; `daemonPID` rejects `pid <= 0`.

## Chat-history audit fixes
- **Skill-creation misroute** — `build me a skill to …` now hits `self:add-skill` before search triggers (the 2026-05-16 unanswered message).
- **New `digest` builtin** — real Hacker News (last 24h, Algolia) + web sweep, strict only-cite-listed-URLs summary; replaces the boilerplate-returning generic search for digest asks.
- **Invisible error replies** — handler fallbacks + reminder shortcuts now persisted via `KrillAgent.RecordTurn`.
- **No turn deadline** — 10-min backstop on Telegram/Discord; hung provider → late apology, not silence.
- **`yes` as a model name** — `Switch` rejects conversational tokens.
- **Health-warning spam** — rate-limited to state transitions.
- **Overstated dive summaries** — detects prose blockers; reports "completed / blocked or awaiting input".

## Testing
- `go build ./...`, `go vet ./...`, full `go test ./...` green.
- New tests: KrillLM provider resolution, digest routing, digest source fetching (incl. URL-less HN fallback), single-source failure tolerance, all-sources-dead error.
- Verified live: daemon boots on `krilllm | gemma4:12b-mlx` with the jarvis personality; `Get me today AI Digest` returns real fresh headlines through gemma 12b.


srvsngh99:
Added docs to this PR: README, docs/PROVIDERS.md, and docs/INSTALL.md now document KrillLM (gemma 12b) as the default provider, with plain Ollama as the custom-model path. Also removed the fabricated Codex model IDs (`gpt-5.5`…) from PROVIDERS.md to match the code (Codex advertises only `auto`).

Note on release notes: the GitHub release body is auto-generated by GoReleaser from commit messages on `v*` tag push — there is no separate release-notes file. The human-facing notes live in CHANGELOG.md (the `0.1.4` entry in this PR).

srvsngh99:
## Review - PR #41 (round 1)

Ran on head `667a648`: `go build ./...` (clean), `go vet ./...` (clean), and targeted tests `go test ./internal/llm/ ./internal/agent/ ./internal/chat/ ./internal/plugin/ ./internal/config/ ./internal/ollama/` (all pass). Read CLAUDE.md (empty) and the project memory. The KrillLM default, the digest skill, the daemon detach fix, and the chat-reliability batch all hang together well and the core mechanics are sound. A few gaps below before this is ready.

### Substantive issues

1. **README.md:13 - version badge still says 0.1.3.** VERSION, `internal/core/types.go` (Version var), and the CHANGELOG all moved to 0.1.4, but the README badge URL is still `version-0.1.3-blue.svg`. The bar for this repo is that the version is consistent across VERSION / types.go / CHANGELOG / README badge, so this is the one straggler. Fix: bump the badge to `version-0.1.4-blue.svg`.

2. **internal/chat/telegram.go:397,401 and internal/chat/discord.go (same block) - the turn-level fallbacks bypass RecordTurn.** This PR's stated goal is "a failed turn no longer looks like the bot stayed silent," and handler.go was carefully updated so error/krill-fact fallbacks call `recordTurn`. But the two replies generated in the transport layer - the handler-error "hit a reef" message and, more importantly, the new deadline apology - are assigned to `resp` *after* `HandleMessage` returns, so they never go through the handler's recordTurn path and are never persisted. That means the single most visible silent-bot case (a provider that hangs until the 10 min `turnDeadline` fires) still leaves zero trace in durable history and recovery context - exactly the hole the PR set out to close. Fix: thread the recorder down to the transport (or expose a small `RecordOutOfBand` on the handler) and persist the deadline/handler-error apology there, the same way handler.go does for its own fallbacks. The user turn is already saved by the agent at agent.go:323 before it hangs, so only the assistant side needs recording here.

3. **internal/agent/planner.go:268-298 - the new blocker-detection logic has no test.** The headline planner fix (the `blockedPrefixes` list, the `result: ... blocked` line check, the `*_` trimming so `**Result: ... blocked.**` is caught, and the 5 -> 10 line window bump) is the thing that stops the "5/5 succeeded" gaslighting, and it is entirely uncovered. The router and digest changes both got regression tests; this one - which is pure string heuristics and the easiest to silently regress - got none. Fix: add a table test for `detectStepBlocker` covering at minimum the `**Result: research step blocked.**` case from the comment, an `i can't actually ...` opener, a `BLOCKED:`/`NEED_INPUT:` token, and a negative case (a normal step output, plus a paragraph that merely quotes a refusal past line 10) so the anchoring guarantee is locked in.

### Nits

- **internal/agent/router.go:288** - `skillCreationPattern` (`\b(build|create|make|add|write)\b[^.!?]*\bskill\b`) is broad because of the greedy `[^.!?]*` between verb and "skill". A genuine memory/chat line like "remember I want to build a skill in chess someday" would now route to `self:add-skill` instead of memory store. All the new test cases are real creation asks, so this is a precision call, not a bug - just flagging the false-positive surface.
- **internal/plugin/digest.go:1315 (Execute) - misleading empty message on a half-dead source.** If `searchFn` is nil and `fetchHN` errors, the combined-error guard is skipped (webErr is nil) and the function returns "sources were reachable but empty" even though HN actually errored. Production always sets `searchFn`, so this is only reachable in a hand-built skill, but the message is wrong for that path.
- **cmd/minikrill/commands.go** - dropping the noiseFilter is the right call for the SIGPIPE fix, but it also drops the second pattern it used to strip (`objc[...]: Class ... implemented in both`), which the `MallocStackLogging=0` env does not cover. Those lines can now reach dive.log. Cosmetic only.

### Verified good

- Daemon detach fix is correct: `out` is an `*os.File`, so `child.Stdout = out` passes the fd directly (no parent-side pipe/goroutine to collapse on Release). Pid is captured before `Release()` and `daemonPID` now rejects `pid <= 0`, closing the "signal every process" footgun.
- KrillLM provider, `NewProvider` aliases, `ResolveTarget` (incl. `gemma12b` shorthand winning over the raw Ollama map), the `IsLocalProvider` gate, and `ListProviders` ordering are all consistent and covered by krilllm_test.go. The stored provider is always the canonical `krilllm`, so the `IsLocalProvider` set not listing the `krill` short alias is fine.
- `RecordTurn` locking is safe - `appendMessage`/`saveTurn` do not re-acquire `a.mu`, so no deadlock - and on the normal path the user turn is saved at agent.go:323 before any error return, so handler.go's assistant-only fallback recordings pair correctly.
- Router ordering is right: `detectSelfSkillTrigger` (incl. the new pattern) runs ahead of the `digest` tool check, so "build me a skill to fetch AI digest" correctly beats the bare-`digest` route. Algolia `numericFilters` `>` is percent-encoded via `url.Values.Encode()`, HN body is `LimitReader`-capped, response body is closed. The unmanaged-Ollama warn is now correctly edge-triggered with a recovery log.
- CHANGELOG is reverse-chronological and consistent with VERSION/types.go. No secrets, no `--no-verify`, no AI attribution in the diff.

Verdict: changes requested - items 1-3 above before merge.

srvsngh99:
Addressed all three substantive items from round 1 in `cb12ec3`:

1. README version badge bumped 0.1.3 -> 0.1.4 to match VERSION, internal/core/types.go, and CHANGELOG.
2. The turn-deadline apology and handler-error fallback now leave a durable-history trace. The deadline wording is centralised in the handler (internal/chat/handler.go), which records it once and is the single user-facing string; the Telegram/Discord deadline overrides that previously replaced an already-recorded fallback are removed, so history no longer disagrees with what the user saw. The exceptional handler-returns-error path persists its last-resort reply via an exported RecordTurn.
3. Added internal/agent/planner_test.go covering detectStepBlocker: structured tags, need_input prose, the new prose blockers (Result: ... blocked, can't actually, same blockers as), markdown-wrapped verdicts, and genuine successes that must not be misclassified.

go build, go vet, and the targeted package tests pass.

srvsngh99:
## Review - PR #41 (round 2)

Ran on head `cb12ec3` (worktree on `feat/krilllm-provider`, rebased on `origin/main`):

- `go build ./...` — clean
- `go vet ./...` — clean
- `go test ./internal/llm/ ./internal/agent/ ./internal/chat/ ./internal/plugin/ ./internal/config/ ./internal/ollama/` — all pass (ollama has no test files)
- Version: VERSION, `internal/core/types.go:13`, CHANGELOG `## [0.1.4]`, and README badge `README.md:13` are all `0.1.4` — consistent
- No secrets, no `--no-verify`, no Co-Authored-By / AI attribution in the commit range

### Round-1 items — all three verified fixed

1. **README version badge** — now `0.1.4` at `README.md:13`. Fixed.

2. **Deadline/handler-error fallbacks bypassing RecordTurn** — Fixed and the design holds up under scrutiny:
   - The deadline wording is centralised in `internal/chat/handler.go:77-89`, gated on `ctx.Err() == context.DeadlineExceeded`, and the same string is both returned to the user AND persisted via `h.recordTurn(...)` at `handler.go:88`. The transports no longer override it. Sent == recorded; no drift between the real deadline string and what's stored.
   - **No double-recording:** `ChatHandlerImpl.HandleMessage` returns `nil` error on every path (reminder, agent-error fallback, empty-fact, normal), so the transports' `RecordTurn` blocks (`telegram.go:403-405`, `discord.go` ~233) only fire on a non-nil error and are effectively defensive dead code with the real handler — they never co-fire with the handler's own `recordTurn`.
   - **No missing user turn:** the user message is persisted by the agent at `internal/agent/agent.go:323-324` (Phase 2), *before* the LLM call that can hang in Phase 3/4. So on a deadline the durable history correctly shows user-msg + assistant-fallback.
   - Exported `RecordTurn` on `*ChatHandlerImpl` (`handler.go:888`) and `KrillAgent.RecordTurn` (`agent.go:532`); the transport type-assertion `interface{ RecordTurn(role, content string) }` matches the pointer-receiver method set (build confirms).

3. **detectStepBlocker had no coverage** — `internal/agent/planner_test.go` added with prose-blocker, `Result: … blocked`, markdown-wrapped, and negative (deep-in-text, genuine-success) cases. Fixed.

### New-regression hunt (clean)

- **No unused imports** after removing the override blocks — `fmt` and `context` are both still used throughout `telegram.go` and `discord.go`.
- **Deadline actually produces an error:** `OllamaProvider.Chat` uses `http.NewRequestWithContext` (`internal/llm/ollama.go:141`), so a `turnDeadline` expiry surfaces as a non-nil error up through `agent.Chat`, triggering the handler's error branch. The `ctx.Err()` check is correct — `msgCtx` is the deadline context, so `ctx.Err()` returns exactly `context.DeadlineExceeded` (not a wrapped `*url.Error` that would need `errors.Is`).

## Substantive issues

1. **Over-broad digest routing matches unrelated substrings** — `internal/agent/router.go:245`: `if strings.Contains(lower, "digest")` routes *any* message containing the substring "digest" to the digest tool. That includes "digesting", "digestive", and "can you digest this document for me" — none of which are AI-news-digest requests. This is a raw substring match sitting directly next to the new word-boundary-aware `skillCreationPattern` regex, so the inconsistency is glaring. Concrete fix: gate it on a word boundary and an AI/news context, e.g. a regex like `\bdigest\b` combined with the existing freshness/AI cues, or at minimum `regexp.MustCompile(`(?i)\bdigest\b`)` so "digesting"/"digestive" don't trip it. The skill-not-found path falls back to chat gracefully (`agent.go:1261-1264`), so this misroutes rather than crashes — but a "digest this PDF" ask will silently hit the HN/AI-news skill instead of the user's intent.

### Nits

- `internal/llm/krilllm.go` / `config/default.yaml`: the primary model id `gemma4:12b-mlx` is unusual — `gemma4` and an `-mlx` suffix aren't standard Ollama tags (current public Gemma is `gemma3`/`gemma2`, and `-mlx` is an Apple MLX convention). If this is a deliberately custom/local tag that's fine, but confirm `ollama pull gemma4:12b-mlx` actually resolves on a clean machine — otherwise `init`'s auto-pull and the documented `minikrill ollama pull gemma4:12b-mlx` will fail for new users.
- `internal/llm/manager.go`: `ResolveTarget` adds `"gemma 12b"` (with a space) as a krill model alias, but `parts[0]` from `/use` splits on whitespace, so a user typing `/use gemma 12b` only ever passes `gemma` to ResolveTarget — the spaced alias is unreachable via the chat command. Harmless, just dead.
- `digest.go:367`: web sweep query is hardcoded to `"AI news today"`; the skill ignores its `input` argument entirely. Fine for an AI-digest builtin, but the name "digest" plus an ignored input could surprise someone asking to "digest <topic>".

### Verified good

- KrillLM provider: `NewProvider` aliases (`krilllm`/`krill-lm`/`krill_lm`), `auto`/empty → primary model, model override all covered by `krilllm_test.go`; embeds `*OllamaProvider` and only overrides `Name()`.
- `IsLocalProvider` correctly gates Ollama auto-start/health-monitoring for both `krilllm` and `ollama` (`cmd/minikrill/commands.go:386`).
- Daemon fixes are sound: pid captured before `Release()` (`commands.go`), `daemonPID` rejects `pid <= 0` guarding the signal-everything footgun, and child inherits the log fd directly instead of an in-process pipe that collapses on parent exit.
- `Switch` conversational-token rejection (`manager.go`) blocks `yes`/`ok`/etc. as model names.
- Unmanaged-Ollama warning is now edge-triggered (`internal/ollama/manager.go:497-518`), with a recovery "reachable again" log.
- Router ordering correct: self-skill (incl. skill-creation regex) at step 2 beats digest/search at step 4 — both directions covered by `router_test.go`.
- Digest skill: tolerates one dead source, errors only when both die, falls back to HN item link when a story has no URL, strict only-cite-listed-URLs system prompt — all covered by `digest_test.go`.

Verdict: CHANGES_REQUESTED — one substantive item (over-broad `digest` substring routing). Everything else, including all three round-1 fixes, is solid.

srvsngh99:
Addressed the round 2 item in `3ccbd02`:

Replaced the raw `strings.Contains(lower, "digest")` route with `digestRequestPattern` (router.go) — "digest" as a whole word adjacent to a news/AI/freshness cue. "digest this document", "help me digest this article", "digesting", and "digestive" now fall through to normal classification instead of hijacking the AI-news digest skill, while "today's AI digest", "news digest", and "daily digest" still route correctly. Added TestRouterDigestDoesNotHijackUnrelated plus more positive cases.

go build, go vet, and the targeted tests pass.

srvsngh99:
## Review - PR #41 (round 3)

Reviewed at head `3ccbd02`. Ran `go build ./...` (clean), `go vet ./...` (clean), and `go test -count=1 ./internal/llm/ ./internal/agent/ ./internal/chat/ ./internal/plugin/ ./internal/config/` (all pass; `internal/ollama` has no test files). Version is a consistent 0.1.4 across VERSION, internal/core/types.go:13, CHANGELOG, and the README badge (README.md:13). No secrets, no `--no-verify`, no AI attribution in the diff.

This round focused on the round-2 digest-routing fix and a full sweep of everything not yet confirmed clean.

### Substantive issues

None. Re-verified the prior-round fixes and found no regressions, and the new surface area holds up.

### Digest routing (the round-2 fix)

`digestRequestPattern` at internal/agent/router.go:171 is correct. I ran the regex against the full matrix rather than eyeballing it:

- Matches: "today's AI digest", "news digest", "daily digest", "morning digest", "tech news digest please", "digest of today's headlines", "send me the industry digest", "the latest AI digest" — all true.
- Does NOT match: "digest this document", "help me digest this article", "digesting", "digestive", "can you digest this for me", "please digest the meeting notes" — all false. The `\b` boundaries kill the "digesting"/"digestive" stems, and the cue requirement kills the bare-verb cases.

The cue-less "get me the digest" / "show me the digest" correctly fall through to search/chat (router.go:254 misses, then searchTriggers/LLM fallback). That is acceptable: a digest ask with zero news/AI/freshness cue is genuinely ambiguous and the LLM fallback is the right place for it.

One known limitation, not a blocker (see nits): the `.*` between cue and `digest` is unanchored, so a multi-clause message that mentions a news cue in one clause and "digest" as a verb in another ("the AI we discussed is fine, can you digest this PDF") will route to the digest skill. Rare in practice, and the failure mode is a degraded-but-harmless digest reply rather than a crash.

### Nits

1. internal/chat/telegram.go:402 and internal/chat/discord.go:~235 — the transport-side `if err != nil { ... RecordTurn(...) }` blocks are effectively dead: `HandleMessage` now returns `(fallback, nil)` on every path (handler.go:89/103), so it never hands a non-nil error back to the transport. The comments already acknowledge this as a defensive backstop, so it is intentional and harmless — just noting it reads as unreachable today.
2. router.go:171 unanchored `.*` cross-clause match described above.
3. internal/agent/agent.go:213 — `ChatFromPlatform` holds `a.mu` for the whole turn, so under a real provider hang the lock is held up to `turnDeadline` (10m) and a second concurrent message blocks on `sync.Mutex` (no ctx awareness). Pre-existing design, explicitly documented at the method, out of scope for this PR.

### Verified good

- KrillLM provider (internal/llm/krilllm.go) cleanly embeds `*OllamaProvider` and overrides `Name()`; `NewKrillLMProvider` resolves empty/"auto" to `gemma4:12b-mlx`. Wiring in provider.go (case "krilllm"/aliases) and manager.go (ListProviders entry, `ResolveTarget` aliases + gemma12b shorthands that win over the raw Ollama map) is consistent. `config.IsLocalProvider` (config.go:118) includes krilllm so it gets Ollama auto-start.
- `conversationalTokens` guard in manager.go Switch correctly rejects "yes"/"ok"/etc. as model names.
- Digest skill (internal/plugin/digest.go): context-aware HTTP, 10s timeout, `>= 400` status check, `io.LimitReader(1MB)`, encoded numericFilters, graceful partial-failure (both dead -> error; one dead -> still produces output; both empty -> explicit "no fresh items"), strict only-cite-listed-URLs system prompt, and a raw fallback if the LLM summary is empty. Registered in builtins.go and covered by three tests (real sources, one-dead-source, all-dead).
- Daemon detach fix (cmd/minikrill/commands.go): `child.Stdout/Stderr = out` (the log *os.File) instead of the old noiseFilter io.Writer wrapper is the correct root-cause fix for the SIGPIPE death; pid captured before `Release()`; `daemonPID` now rejects `pid <= 0` (guards signalling -1/0). The noiseFilter type is fully removed with no dangling references (build confirms).
- RecordTurn / turn-deadline flow: user turn is persisted at agent.go:324 before any handler runs, and the handler persists the assistant fallback (handler.go:88) on the error/deadline path, so durable history stays consistent. Reminder shortcut persists both sides (handler.go:56-57). No re-entrant `a.mu` deadlock — every `recordTurn` call happens after `ChatFromPlatform` has returned and released the lock. Deadline-aware wording distinguishes timeout from genuine failure.
- Planner blocker detection (planner.go): new `blockedPrefixes` + `Result: … blocked` verdict detection, `*_` trimming, and the 10-line window; the "blocked" status correctly `continue`s past the `done` assignment and the tally counts blocked/need_input separately from done. Covered by planner_test.go (markdown-wrapped result, result verdict, "can't actually", "same blockers").

Verdict: APPROVED — ship it.

srvsngh99:
Approved after 3 automated review rounds. Round 1 fixed the README badge, deadline-apology persistence, and detectStepBlocker test coverage; round 2 tightened digest routing to a whole-word + news-cue pattern; round 3 swept the full diff clean and approved. Merging.

PR #42, opened 2026-06-18
Add Reef web channel (WebBot)

srvsngh99:
## Why

Part of moving the Krill colony off Telegram and onto the Reef hub (the agent-centric messaging app). This PR makes Mini-Krill run headless on Reef as a `core.ChatBot`.

## What changed

- `internal/reef/client.go` (new) - dependency-free Reef client. `PostIngest` posts messages; `PollOutbox` long-polls the owner outbox (ctx-aware, checks HTTP status). Inert unless `REEF_INGEST_URL` + `REEF_AGENT_TOKEN` are set.
- `internal/chat/web.go` (new) - `WebBot` (`Platform()` = `web`): long-polls the outbox and runs each owner message through the shared `ChatHandler`. Dispatches are bounded by a semaphore and each turn runs under `turnDeadline`, so a burst cannot spawn unbounded goroutines and a hung turn cannot leak. On shutdown a bounded `Stop()` drains in-flight turns (the hub marks outbox items delivered on drain and does not redeliver).
- `cmd/minikrill/commands.go` - `startBots` runs web-only when Reef is configured (one adapter at a time), registers a `web` task notifier, and drains the WebBot after `cancel()` in both the dive daemon and TUI shutdown paths.
- `cmd/minikrill/main.go` - fixes a latent `friendlyError(nil)` panic; drops the em-dash from the root command summary.

## How it was tested

- `go build ./...`, `go vet ./...`, `gofmt -l` clean; `go test ./internal/chat/...` and `go test -race ./internal/chat/...` pass.
- Ran `minikrill dive` live against the Reef hub: posts status and answers owner DMs.
- Three adversarial review rounds; round 3 verified the shutdown drain is actually wired in, bounded by `drainGrace`, and free of deadlock or goroutine leak.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,3 +46,4 @@ docs/promo/
 credentials.*
 *.key
 *.pem
+com.deepkrill.minikrill.plist
--- a/cmd/minikrill/commands.go
+++ b/cmd/minikrill/commands.go
@@ -458,6 +458,11 @@ var diveCmd = &cobra.Command{
 		fmt.Println()
 		fmt.Println(cCyan + "  Mini Krill is surfacing..." + cReset)
 		cancel()
+		if bs.web != nil {
+			// Stop the poll loop (via cancel above) then drain in-flight turns
+			// so a reply the hub already marked delivered still posts.
+			_ = bs.web.Stop()
+		}
 		_ = stack.hb.Stop()
 		_ = stack.mcp.Close()
 		return nil
@@ -584,6 +589,7 @@ type botStatus struct {
 	DiscordErr  error
 	WebOK       bool
 	WebErr      error
+	web         *chat.WebBot // set in reef mode; drained on shutdown
 }
 
 // startBots launches Telegram and Discord bots in the background when enabled.
@@ -606,6 +612,7 @@ func startBots(ctx context.Context, stack *krillStack) botStatus {
 			}
 		}()
 		s.WebOK = true
+		s.web = wbBot
 		klog.Info("web (reef) bot started", "agent", reef.AgentID())
 		return s
 	}
@@ -688,13 +695,18 @@ var chatCmd = &cobra.Command{
 		_ = stack.hb.Start(ctx)
 		defer func() { _ = stack.hb.Stop() }()
 
+		var bs botStatus
 		if daemonPID() == 0 {
-			bs := startBots(ctx, stack)
+			bs = startBots(ctx, stack)
 			if stack.cfg.Telegram.Enabled && bs.TelegramErr != nil {
 				klog.Warn("telegram bot failed to start", "error", bs.TelegramErr)
 			}
 		} else {
 			klog.Debug("skipping bot startup, dive daemon is running")
+		}
+		if bs.web != nil {
+			// On TUI exit, cancel the poll loop then drain in-flight turns.
+			defer func() { cancel(); _ = bs.web.Stop() }()
 		}
 
 		app := tui.NewApp(stack.agent, stack.brain, stack.hb, stack.skills, stack.mcp, core.Version, stack.cfg.Log.File)
--- a/cmd/minikrill/commands.go
+++ b/cmd/minikrill/commands.go
@@ -23,6 +23,7 @@ import (
 	klog "github.com/srvsngh99/mini-krill/internal/log"
 	"github.com/srvsngh99/mini-krill/internal/ollama"
 	"github.com/srvsngh99/mini-krill/internal/plugin"
+	"github.com/srvsngh99/mini-krill/internal/reef"
 	"github.com/srvsngh99/mini-krill/internal/reminder"
 	"github.com/srvsngh99/mini-krill/internal/tui"
 )
@@ -415,14 +416,17 @@ var diveCmd = &cobra.Command{
 		fmt.Printf("  "+cDim+"Skills:"+cReset+"   %d registered\n", len(stack.skills.List()))
 
 		bs := startBots(ctx, stack)
-		if stack.cfg.Telegram.Enabled {
+		if bs.WebOK {
+			fmt.Println("  " + cGreen + "Reef:     swimming" + cReset)
+		}
+		if stack.cfg.Telegram.Enabled && !reef.IsConfigured() {
 			if bs.TelegramOK {
 				fmt.Println("  " + cGreen + "Telegram: swimming" + cReset)
 			} else {
 				fmt.Printf("  "+cRed+"Telegram: %s\n"+cReset, friendlyError(bs.TelegramErr))
 			}
 		}
-		if stack.cfg.Discord.Enabled {
+		if stack.cfg.Discord.Enabled && !reef.IsConfigured() {
 			if bs.DiscordOK {
 				fmt.Println("  " + cGreen + "Discord:  swimming" + cReset)
 			} else {
@@ -578,12 +582,33 @@ type botStatus struct {
 	TelegramErr error
 	DiscordOK   bool
 	DiscordErr  error
+	WebOK       bool
+	WebErr      error
 }
 
 // startBots launches Telegram and Discord bots in the background when enabled.
 // Bots are tied to ctx and will shut down when the context is cancelled.
 func startBots(ctx context.Context, stack *krillStack) botStatus {
 	var s botStatus
+	// One adapter at a time: when Reef is configured, run only the web bot.
+	if reef.IsConfigured() {
+		wbBot := chat.NewWebBot(stack.handler)
+		if runner := stack.agent.TaskRunnerRef(); runner != nil {
+			runner.RegisterNotifier("web", func(chatID, message string) {
+				if err := reef.PostIngest("chat", "chat", message); err != nil {
+					klog.Warn("reef notify failed", "error", err)
+				}
+			})
+		}
+		go func() {
+			if err := wbBot.Start(ctx); err != nil {
+				klog.Error("web (reef) error", "error", err)
+			}
+		}()
+		s.WebOK = true
+		klog.Info("web (reef) bot started", "agent", reef.AgentID())
+		return s
+	}
 	if stack.cfg.Telegram.Enabled {
 		tgBot, err := chat.NewTelegramBot(stack.cfg.Telegram, stack.handler)
 		if err != nil {
--- a/cmd/minikrill/main.go
+++ b/cmd/minikrill/main.go
@@ -53,31 +53,29 @@ func main() {
 
 var rootCmd = &cobra.Command{
 	Use:     "minikrill",
-	Short:   "Mini Krill - Your crustaceous AI buddy",
+	Short:   "mini-krill — a lightweight, local-first AI agent",
 	Version: core.Version,
-	Long: cBCyan + `
-   .-''''-.
-  /  >o  /)    Mini Krill
- |  /___/ |    by Sourav Singh / Sourav AI Labs
-  \__\_\_/` + cReset + `
+	Long: "  " + cBold + brand.Wordmark + cReset + "   " + cDim + brand.Tagline + cReset + `
 
   A lightweight, open-source AI agent that runs locally via
   Ollama or through subscription-backed Codex/Claude CLIs.
 
   ` + cDim + `Get started:` + cReset + `
-    ` + cCyan + `minikrill init` + cReset + `       Setup wizard
-    ` + cCyan + `minikrill chat` + cReset + `       Interactive chat
-    ` + cCyan + `minikrill dive` + cReset + `       Start background services
-    ` + cCyan + `minikrill tui` + cReset + `        Terminal dashboard
-    ` + cCyan + `minikrill doctor` + cReset + `     Health diagnostics
+    ` + cBold + `minikrill init` + cReset + `       Setup wizard
+    ` + cBold + `minikrill chat` + cReset + `       Interactive chat
+    ` + cBold + `minikrill dive` + cReset + `       Start background services
+    ` + cBold + `minikrill tui` + cReset + `        Terminal dashboard
+    ` + cBold + `minikrill doctor` + cReset + `     Health diagnostics
 
   ` + cDim + `Documentation:` + cReset + `
-    ` + cCyan + `README.md` + cReset + `            Overview and usage
-    ` + cCyan + `docs/INSTALL.md` + cReset + `     Install and setup
-    ` + cCyan + `docs/PROVIDERS.md` + cReset + `   Ollama, Codex, Claude
-    ` + cCyan + `docs/MEMORY.md` + cReset + `      Memory and preferences
-    ` + cCyan + `docs/INTERFACES.md` + cReset + `  CLI, Telegram, Discord
-    ` + cCyan + `docs/TESTING.md` + cReset + `     Feature test checklist`,
+    ` + cBold + `README.md` + cReset + `            Overview and usage
+    ` + cBold + `docs/INSTALL.md` + cReset + `     Install and setup
+    ` + cBold + `docs/PROVIDERS.md` + cReset + `   Ollama, Codex, Claude
+    ` + cBold + `docs/MEMORY.md` + cReset + `      Memory and preferences
+    ` + cBold + `docs/INTERFACES.md` + cReset + `  CLI, Telegram, Discord
+    ` + cBold + `docs/TESTING.md` + cReset + `     Feature test checklist
+
+  ` + cDim + brand.LabMark + "  " + brand.Lab + "  ·  " + brand.Site + cReset,
 }
 
 func init() {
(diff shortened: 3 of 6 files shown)


srvsngh99:
## Review - PR #42 (round 1)

Ran on `feat/reef-channel` @ `b1eb923`: `go build ./...` (pass), `go vet ./...` (clean), `gofmt -l` on the four changed .go files (empty/clean), `go test ./internal/reef/... ./internal/chat/...` (reef has no tests, chat ok). Cross-checked the WebBot/notifier wiring against the Telegram bot and the live Reef server contract in `~/reef/app.py`.

### Substantive issues

1. **`cmd/minikrill/main.go:56` - em-dash in a user-facing string (house-rule violation).** `Short: "mini-krill — a lightweight, local-first AI agent"` ships a literal U+2014 in the CLI help output. We use hyphens, never em/en dashes, in code strings and comments. This is the only new dash the PR introduces. Fix: replace the em-dash with `-` (or `--`): `"mini-krill - a lightweight, local-first AI agent"`.

2. **`internal/chat/web.go:54,65` - `dispatch` runs the agent turn on the raw bot context with no per-turn deadline; a hung turn leaks a goroutine forever.** The Telegram path wraps every `HandleMessage` in `context.WithTimeout(ctx, turnDeadline)` (`internal/chat/telegram.go:382`) precisely so a stuck plan/execute cycle can't pin a goroutine. WebBot passes `ctx` straight through and fires each turn with `go w.dispatch(ctx, text)`, so a turn that never returns is an unbounded leak, and there is no upper bound on how many `dispatch` goroutines can be in flight at once (one per outbox item, every poll). Fix: in `dispatch`, derive `msgCtx, cancel := context.WithTimeout(ctx, <turnDeadline>)` / `defer cancel()` and pass `msgCtx` to `HandleMessage`; consider a small worker-pool or semaphore to cap concurrent turns the way `TaskRunner.maxConcurrent` does.

3. **`internal/chat/web.go:31,54` - in-flight messages are silently lost on shutdown because the Reef outbox marks items delivered at drain time, not at ack time.** `drain_outbox` in `~/reef/app.py:198` flips `delivered=true` the instant `PollOutbox` returns, so there is no server-side redelivery. But the owner message is only acted on inside the fire-and-forget `go w.dispatch`. When `ctx` is cancelled, `Start` returns immediately (web.go:39) while any dispatched-but-unfinished turns are abandoned mid-flight - the message is already marked delivered and will never come back. Same loss if the process is killed between drain and reply. At minimum, on `ctx.Done()` wait for in-flight dispatches (a `sync.WaitGroup` drained before `Start` returns) so a graceful shutdown finishes what it pulled; document that hard kills can drop the last batch.

4. **`internal/reef/client.go:80-91` - `PollOutbox` never checks the HTTP status code before decoding.** `PostIngest` correctly rejects `>= 300` (client.go:61), but the poll path goes straight from `Do` to `json.NewDecoder(resp.Body).Decode`. A 401 (bad/rotated token) or 5xx returns a non-JSON body, the decode fails, and the bot reports a generic decode error and silently 5-second-backoff-loops forever with no signal that auth is actually broken. Fix: after the `Do`, `if resp.StatusCode >= 300 { return nil, fmt.Errorf("reef outbox: status %d", resp.StatusCode) }` before decoding, mirroring `PostIngest`.

### Nits

- `internal/reef/client.go:55,85` - a fresh `http.Client` is allocated per call, so keep-alive connection reuse is lost on a loop that polls every ~25s. Hoist a single `var httpClient = &http.Client{...}` (or two, given the differing timeouts) at package scope.
- `internal/reef/client.go:49,80` - use `http.NewRequestWithContext` so request cancellation tracks the caller's context; today neither POST nor GET is cancellable mid-flight even when the bot ctx is done.
- `internal/reef/client.go:58` - the `io.Copy(io.Discard, resp.Body)` return value is dropped; fine functionally, but `_ =` it for vet-friendliness / intent.
- `internal/chat/web.go` duplicates the `log.Info("web (reef) bot started", ...)` line that `startBots` (commands.go:609) already emits; pick one site so the log isn't doubled on every launch.
- `internal/chat/web.go:60` - `dispatch` ignores `it.ID`/`it.Type`; if Reef ever sends non-chat outbox types (it already emits `callback` items, app.py:896) they'll be run through the chat handler as if they were prose. Consider filtering on `it.Type == "reply"`/owner text.

### Verified good

- `core.ChatBot` is fully and correctly implemented: `Start(ctx) error`, `Stop() error`, `Platform() string` all present and matching `internal/core/types.go:328`.
- The `friendlyError(nil)` fix (`main.go:250`) is correct - it returns `"not started"` and prevents the nil `err.Error()` deref; the old code would panic.
- HTTP bodies are closed via `defer resp.Body.Close()` in both calls; `PostIngest` drains the body before close so the connection can be reused.
- No hardcoded secrets: the token is read from `REEF_AGENT_TOKEN`; the filled plist is gitignored and only a sanitized `*.template` is committed.
- One-adapter-at-a-time gating reads correctly: `startBots` returns early in reef mode and the dive status block gates Telegram/Discord on `!reef.IsConfigured()`.
- Branch is `feat/reef-channel` (not main/master); no `--no-verify`; no AI attribution in the new files.

### Verdict

CHANGES_REQUESTED - the em-dash and the unbounded/uncancellable dispatch goroutines (plus the drain-time message-loss window and the missing poll status check) need to land before this merges.

srvsngh99:
Addressed all four substantive items in 43ddf9c.

1. Em-dash in Short string - fixed (hyphen).
2. Unbounded goroutines / no per-turn timeout - fixed. Dispatches are bounded by a semaphore (maxConcurrentTurns) and each turn runs under a turnDeadline timeout, matching the telegram and discord paths, so a hung turn cannot leak forever.
3. In-flight dispatches lost on shutdown - fixed. A WaitGroup tracks dispatches and Start drains them on ctx cancel; the turn body is detached from the poll ctx (capped by turnDeadline) so a reply the hub already marked delivered still posts during shutdown drain. The hub does not redeliver, so this avoids silent loss.
4. PollOutbox no HTTP status check - fixed. It now returns an error on status >= 300 and is ctx-aware (http.NewRequestWithContext), so a 401/5xx is logged rather than retried silently.

srvsngh99:
## Review - PR #42 (round 2)

Ran on head `43ddf9c`: `go build ./...` clean, `go vet ./...` clean, `gofmt -l` on the 4 changed .go files empty, `go test ./internal/reef/... ./internal/chat/...` green (reef has no test files). No secrets in the diff (plist ships a `__REEF_AGENT_TOKEN__` placeholder and the filled-in plist is gitignored - good), no `--no-verify`, no em/en dashes in added lines, no AI attribution.

Re-checking the 4 round-1 items:

1. Em-dash in the `Short` string - fixed. `cmd/minikrill/main.go:56` now reads `mini-krill - a lightweight, local-first AI agent` with an ASCII hyphen.
2. Per-turn timeout + unbounded `go w.dispatch` - fixed. `internal/chat/web.go:36` adds `sem chan struct{}` capped at `maxConcurrentTurns = 4`, the loop acquires a slot before spawning (`web.go:71-75`), and each turn body is wrapped in `context.WithTimeout(..., turnDeadline)` (`web.go:83`). Concurrency is bounded and a hung turn can no longer leak forever.
4. `PollOutbox` HTTP status + ctx-awareness - fixed. `internal/reef/client.go:84` uses `http.NewRequestWithContext`, and `client.go:88-91` returns `fmt.Errorf("reef outbox: status %d", ...)` on `StatusCode >= 300` after draining the body. `PostIngest` got the same treatment (`client.go:58-60`).

Item 3 (drain via WaitGroup) is where the design does not hold up - see issue 1 below.

### Issues

1. **The graceful drain is dead code - nothing in `dive` waits for the WebBot goroutine, so replies the hub already marked delivered are still dropped on shutdown.** `startBots` launches `go wbBot.Start(ctx)` (`cmd/minikrill/commands.go:594-598`) and returns; no reference to `wbBot` is kept and `wbBot.Stop()` is never called. On SIGTERM the dive handler does `cancel()` then `_ = stack.hb.Stop()`, `_ = stack.mcp.Close()`, `return nil` (`commands.go:456-463`) - it never joins the WebBot goroutine. So when `cancel()` fires, `Start` returns and its `defer w.wg.Wait()` (`web.go:46`) runs *inside the detached goroutine*, racing the process exit that `return nil` triggers. The OS reaps the process before the drain completes. Net effect: the WaitGroup + detached-`context.Background()` turn context add real complexity but deliver none of their stated benefit - the exact data-loss the comment at `web.go:67-70` says it prevents can still happen.

   This also resolves the launchd concern raised for round 2 the wrong way: shutdown does NOT block up to `turnDeadline` (good for launchd SIGTERM), but only because the drain is ineffective, not because it was designed to bound the wait. The comment at `web.go:80-82` ("so an in-flight reply still posts during shutdown drain") is misleading - there is no drain that the process actually observes.

   Fix: make the lifecycle real and bounded. Capture the bot and join it on shutdown with a cap, e.g. in `startBots` return the `*WebBot` (or store it on the stack), and in the dive handler after `cancel()`:
   ```go
   done := make(chan struct{})
   go func() { _ = wbBot.Stop(); close(done) }()
   select {
   case <-done:
   case <-time.After(5 * time.Second): // bound the drain for launchd
       klog.Warn("web bot drain timed out")
   }
   ```
   With a bounded join the detached `context.Background()` turn context is justified; without a join, drop the WaitGroup/detached-ctx machinery entirely and just let the poll ctx cancel the turn, since the drain buys nothing.

### Nits

- `web.go:46` `defer w.wg.Wait()` and `Stop()` (`web.go:112`) both call `wg.Wait()`; harmless (idempotent) but only one is reachable today since `Stop` is never invoked. Once issue 1 is addressed, prefer driving the drain from `Stop()` only and drop the `defer` in `Start`, so there is a single owner of the wait.
- `commands.go` registers the `web` notifier (`PostIngest("chat","chat",message)`) for async task completions, which posts to the same Reef chat channel as the synchronous `dispatch` reply. Different code paths so no double-send today, but worth a one-line comment noting the two posting paths share the channel.
- `web.go:48` `if ctx.Err() != nil { return nil }` at the top of the loop is redundant with the `ctx.Done()` checks inside `PollOutbox` and the `select`; not wrong, just belt-and-suspenders.

### Verified good

- Semaphore / WaitGroup / ctx interplay has no deadlock: the loop is the only producer on `w.sem`, and once `Start` returns no goroutine blocks on `w.sem <-`; in-flight goroutines each `<-w.sem` and `wg.Done()` independently.
- `sleepCtx` (`web.go:116-124`) is correct - timer is stopped via defer, returns false on ctx cancel.
- Detached turn context is correctly capped at `turnDeadline` (10m, `handler.go:23`), matching telegram/discord parity, so a turn cannot leak forever even with `context.Background()`.
- `PollOutbox` client timeout `waitSeconds+10` (`client.go:85`) correctly exceeds the 25s long-poll window; ctx still aborts the request early on shutdown.
- Adapter exclusivity: when Reef is configured, telegram/discord are suppressed both in startup (`commands.go`) and status output (`commands.go:422,429`). Clean.
- Plist template secret handling is correct: placeholder token, gitignored real file, `chmod 600` instructions.

Verdict: Changes requested - item 3's drain does not actually run on shutdown; wire the WebBot into the dive lifecycle with a bounded join (or remove the drain machinery as unused).

srvsngh99:
Addressed the round-2 item in e97da56.

1. Drain never ran on shutdown - fixed. The WebBot is now threaded through botStatus (commands.go startBots sets s.web), and the surface path calls bs.web.Stop() right after cancel() in both the dive daemon path (commands.go ~460) and the TUI path (deferred). Stop() is bounded by drainGrace (5s) via a select on the WaitGroup, so it gives in-flight turns a short window to post the reply the hub already marked delivered without holding launchd's SIGTERM window open. The previously-unreachable `defer wg.Wait()` inside Start is removed since the join now happens in the shutdown path. go build / vet / gofmt clean; go test -race ./internal/chat/... passes.

srvsngh99:
## Review - PR #42 (round 3)

Ran on head `e97da56`: `go build ./...` clean, `go vet ./...` clean, `gofmt -l` on the changed .go files empty, `go test ./internal/reef/... ./internal/chat/...` green (reef still has no test files), `go test -race ./internal/chat/...` green (uncached). No em/en dashes in added lines, no secrets in the diff (plist ships the `__REEF_AGENT_TOKEN__` placeholder and the filled-in file is gitignored), no `--no-verify`, no AI attribution.

This round was about the round-2 finding: the graceful drain was dead code because nothing joined the WebBot goroutine on shutdown. That is now fixed and the fix holds up.

### Round-2 item 3 - verified fixed

- (a) Daemon path drains. `cmd/minikrill/commands.go:460` calls `cancel()`, then `461-465` does `if bs.web != nil { _ = bs.web.Stop() }`. The WebBot is now captured on `botStatus.web` (`commands.go:592`, set in `startBots` at `615`), so the join actually runs on SIGTERM instead of racing process exit.
- (b) TUI path drains. `commands.go:707-710` registers `defer func() { cancel(); _ = bs.web.Stop() }()` when `bs.web != nil`.
- (c) `Stop()` is bounded and cannot block past `drainGrace`. `internal/chat/web.go:119-128` spawns `go { wg.Wait(); close(done) }` and selects `done` against `time.After(drainGrace)` (5s, `web.go:23`). Worst case it returns after 5s and logs the timeout - launchd's SIGTERM window is never held hostage.
- (d) No deadlock or leak across the bounded Stop, the detached turn ctx, the semaphore, and ctx-cancel. After `cancel()` the poll loop is the only producer on `w.sem` and it has returned (`web.go:53/58/78`), so nothing blocks on `w.sem <-` anymore; each in-flight goroutine independently does `<-w.sem` and `wg.Done()` (`web.go:83-84`). The throwaway waiter goroutine in `Stop` only blocks on `wg.Wait()` and is itself bounded by `turnDeadline`, then the process exits.
- (e) Detached turns are still capped. `web.go:88` wraps each turn in `context.WithTimeout(context.Background(), turnDeadline)` (10m, `handler.go:23`), so a hung turn cannot leak forever even though it is detached from the poll ctx. The previously-unreachable `defer w.wg.Wait()` in `Start` has been removed, so the drain has a single owner (`Stop`).

### Round-1 items 1, 2, 4 - no regression

1. Em-dash in `Short` - still ASCII hyphen (`main.go`).
2. Per-turn timeout + bounded concurrency - intact. `web.go:37` sem capped at `maxConcurrentTurns = 4`, slot acquired before spawn (`web.go:76-80`), each turn wrapped in `turnDeadline` (`web.go:88`).
4. `PollOutbox` status check + ctx-awareness - intact. `internal/reef/client.go` uses `http.NewRequestWithContext` and returns an error on `StatusCode >= 300` before decoding.

### New-regression hunt on e97da56

Checked the TUI `var bs botStatus` hoist and the deferred `cancel(); Stop()` against the existing `defer cancel()` at `commands.go:694`:

- Defer ordering is correct. LIFO on return from `app.Run()`: the web-drain defer (`709`) runs first (cancel poll ctx, then drain), then `hb.Stop` (`696`), then the outer `cancel()` (`694`). The drain runs before the outer cancel, which is what we want.
- Calling `cancel()` twice is safe - `context.CancelFunc` is idempotent, the second call is a no-op.
- `bs.web` is nil-safe in both paths (guarded by `if bs.web != nil`), and when a dive daemon is already running `startBots` is skipped so `bs` stays zero-valued and the drain defer is never registered.

No new bug found.

### Nits

- `botStatus.WebErr` (`commands.go:591`) is declared but never set or read - `startBots` only ever sets `WebOK = true` in reef mode. It parallels `TelegramErr`/`DiscordErr`, so harmless, but it is dead today; either populate it on a start failure or drop it.
- Still no test coverage for `WebBot` (Start/Stop/drain) or the `reef` client. The drain/semaphore/detached-ctx interplay is exactly the kind of concurrency code worth a small table test (e.g. a fake handler that blocks, assert `Stop` returns within ~drainGrace and the reply still posts). Not blocking, but this logic has churned three rounds with zero tests guarding it.
- The `go func() { w.wg.Wait(); close(done) }()` waiter in `Stop` outlives `Stop` if the drain times out. It is bounded by `turnDeadline` and the process is exiting, so benign - just noting it is an intentional detached waiter, not a leak.

### Verified good

- All gates clean (build, vet, gofmt, test, race).
- Stop is bounded by `drainGrace` and cannot block past 5s.
- Detached turn ctx capped by `turnDeadline`; no unbounded goroutine leak.
- Semaphore/WaitGroup/ctx interplay is deadlock-free; the unreachable `defer wg.Wait()` from round 2 is gone.
- Adapter exclusivity still holds: reef mode suppresses telegram/discord in both startup and status output (`commands.go:422,429`).
- Plist secret handling correct: placeholder token, gitignored real file, `chmod 600` instructions.
- No dashes in added lines, no secrets, no `--no-verify`, no AI attribution.

Verdict: APPROVED - the drain now actually runs on shutdown in both paths and is properly bounded; items 1, 2, 4 are intact; no new regression. The nits (dead `WebErr` field, missing tests) are non-blocking follow-ups.

srvsngh99:
Approved after 3 automated review rounds. Merging.

PR #43, opened 2026-06-19
Reef WebBot: at-least-once delivery (lease, ack, dedup)

srvsngh99:
## Why
The Reef hub now supports at-least-once delivery (lease + ack), but Mini-Krill's WebBot still consumed outbox items on pull (at-most-once), so a crash between pulling an owner message and posting the reply silently dropped it.

## What changed
- `internal/reef/client.go`: `PollOutbox(ctx, wait, ack)` adds `&ack=1`; new `Ack(ctx, ids)` POSTs `/api/ack`.
- `internal/chat/web.go`: the WebBot polls with `ack=true`, acks each item inside its dispatch goroutine once the turn finishes (detached + 15s timeout so it still completes during shutdown drain), consumes empty items, and dedupes redeliveries via a bounded `seenSet` (touched only by the single Start goroutine). An already-seen redelivery is re-acked so a lost ack self-heals.
- `seenset_test.go`: unit tests for dedup, empty-id, FIFO eviction.

## How it was tested
- `go build ./...`, `go vet`, `go test ./internal/chat -run TestSeenSet` all pass. Hub side already merged + deployed; legacy and ack pollers coexist, so this rolls out independently.

### DIFF
--- a/internal/chat/seenset_test.go
+++ b/internal/chat/seenset_test.go
@@ -1,0 +1,34 @@
+package chat
+
+import "testing"
+
+func TestSeenSetDedup(t *testing.T) {
+	s := newSeenSet(3)
+	if !s.take("a") {
+		t.Fatal("first take of a should be new")
+	}
+	if s.take("a") {
+		t.Fatal("second take of a should be a duplicate")
+	}
+}
+
+func TestSeenSetEmptyIDNeverDedups(t *testing.T) {
+	s := newSeenSet(3)
+	if !s.take("") || !s.take("") {
+		t.Fatal("empty id must always be treated as new (never deduped)")
+	}
+}
+
+func TestSeenSetEviction(t *testing.T) {
+	s := newSeenSet(3)
+	s.take("a")
+	s.take("b")
+	s.take("c")
+	s.take("d") // exceeds max 3 -> oldest ("a") evicted
+	if s.take("b") {
+		t.Fatal("b should still be remembered (within window)")
+	}
+	if !s.take("a") {
+		t.Fatal("a should have been evicted and read as new again")
+	}
+}
--- a/internal/chat/web.go
+++ b/internal/chat/web.go
@@ -41,19 +41,23 @@ func NewWebBot(handler core.ChatHandler) *WebBot {
 func (w *WebBot) Platform() string { return "web" }
 
 // Start posts a presence ping then long-polls the Reef outbox until ctx is
-// cancelled, dispatching each owner message to the handler. On shutdown it
-// waits for in-flight dispatches to finish so replies the hub already marked
-// delivered are not silently dropped (the hub does not redeliver).
+// cancelled, dispatching each owner message to the handler. Delivery is
+// at-least-once: it polls with ack=1 (the hub leases items rather than consuming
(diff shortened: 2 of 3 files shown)


srvsngh99:
## Review - PR #43

Pulled `feat/reef-ack`, read the full diff plus `web.go` and `reef/client.go`. Opting the WebBot into the hub's at-least-once delivery. I went in assuming the concurrency was wrong and tried to break it; it holds up.

`go test ./internal/chat/... ./internal/reef/...` passes, including `go test -race ./internal/chat/...` (clean, no data race).

### Substantive issues

none.

I traced the three concerns hardest:

- **seenSet race.** `seen.take` is called at `web.go:74`, inside the single Start loop, never inside a dispatch goroutine. The set is never passed into a goroutine; goroutines only capture `id` by value. No shared mutable access, so the lockless comment is honest. `-race` agrees.
- **Every polled item gets acked.** All three paths are covered: already-seen redelivery re-acks and continues (`web.go:74-77`), empty text acks and continues (`web.go:79-81`), normal item acks in the goroutine after dispatch (`web.go:105`). The only unacked path is shutdown while blocked on the semaphore (`web.go:88-89`), where the item stays leased and the hub redelivers next run. Nothing falls through to redeliver-forever.
- **Ack on shutdown / leak safety.** `defer w.wg.Done()` and `defer func(){ <-w.sem }()` are both at the top of the goroutine, so the WaitGroup and semaphore are always released even if `dispatch` panics. The ack uses a detached `context.Background()` so request-ctx cancellation does not abort it, and `Stop` drains in-flight goroutines via `wg.Wait`. An ack slower than `drainGrace` (5s) gets abandoned at exit; acceptable since the lease redelivers.

### Nits

1. `seenSet` eviction is correct (holds exactly `max` ids, FIFO; the eviction test pins this), but with `delete` + `s.order = s.order[1:]` the backing array never shrinks and old strings under the slice header stay reachable until enough appends reallocate. At max=1000 this is nothing, just noting it is a slow-growing slice header, not a real leak.
2. The dedup re-ack at `web.go:75` uses the request `ctx` while the post-dispatch ack at `web.go:105` uses a detached ctx. Intentional and documented, fine, just an asymmetry a future reader may trip on.

### Pre-existing (not this PR, but heads up)

`go build ./...` and `go vet ./...` fail on `cmd/minikrill/main.go` (`undefined: brand.Wordmark`, `brand.LabMark`, `brand.Lab`, `brand.Site`). This reproduces on `main` too, so it is not introduced here, but the tree does not currently build end-to-end. Worth fixing on a separate branch.

### Verified good

- `&ack=1` query build is valid; `ack=false` keeps the legacy consuming path.
- `Ack` guards `!IsConfigured()` and empty ids, sets the token + content-type, drains and closes the body, and maps status >= 300 to an error.
- `PollOutbox` signature change has exactly one caller (`web.go:60`); no other caller or test broke.
- Commit message is clean: no AI attribution, no secrets, no em/en dashes in added lines.

Verdict: approve. Single-goroutine seenSet, all ack paths covered, leak-safe defers, detached shutdown ack. Ship it.

srvsngh99:
Approved after review (race-clean dispatch, every item acked, no goroutine/semaphore leak, lost-ack self-heal). Merging.

PR #48, opened 2026-06-25
FeedWatcher: genuine appraisal-driven feed engagement

srvsngh99:
## What the user asked for

> how can we enable mini rill and lab krill running on local krill using gemma 12b model to interact with these feeds ... i want geninue interaction comments likes replies on comments just like social media for agents

This is the Mini-Krill half of that: a genuine, selective feed presence (not forced engagement).

## What changed

A `FeedWatcher` goroutine runs beside the existing Reef DM loop. It observes every post but acts on few: whether to like, comment, or reply is an appraisal run through the local model (gemma4:12b-mlx) primed with the agent's identity and memory. Three levers keep it genuine rather than spammy:
- a relevance threshold (only engage when truly interested);
- a per-hour attention budget (a hard sliding-window cap, forces choosing);
- an agent-to-agent ping-pong guard (reply chains decay as the bar rises with depth, hard stop at MaxReplyDepth, an owner reply resets the count).

Every engagement is written to brain memory for continuity and dedup.

- `internal/reef`: GetFeed, GetComments, PostComment, LikePost, LikeComment.
- `internal/feed`: the watcher plus appraisal/budget/ping-pong logic, unit tested.
- `config`: opt-in Feed block (Mini-Krill tuned lively: threshold 0.55, budget 12); wired into startBots behind reef.IsConfigured().

Round 1 review fix: `agentChainDepth` now guards against a cyclic comment chain from untrusted hub data that could hang the watcher goroutine.

Requires the Reef accounts/bridge work deployed and a per-account REEF_AGENT_TOKEN before it can actually comment (comments are identity bound server side).

## How it was tested

`go build ./...`, `go vet`, `go test -race ./internal/feed/ ./internal/reef/ ./internal/config/` all pass (unit tests cover appraisal JSON parsing, budget cap, ping-pong depth including the cycle guard). Two adversarial Opus review rounds.

## Follow-up (non blocking)

A memory-write failure right after a successful post could cause a re-appraise next cycle; bounded by the budget. Worth a warn log later.

### DIFF
--- a/cmd/minikrill/commands.go
+++ b/cmd/minikrill/commands.go
@@ -19,6 +19,7 @@ import (
 	"github.com/srvsngh99/mini-krill/internal/config"
 	"github.com/srvsngh99/mini-krill/internal/core"
 	"github.com/srvsngh99/mini-krill/internal/doctor"
+	"github.com/srvsngh99/mini-krill/internal/feed"
 	"github.com/srvsngh99/mini-krill/internal/llm"
 	klog "github.com/srvsngh99/mini-krill/internal/log"
 	"github.com/srvsngh99/mini-krill/internal/ollama"
@@ -611,6 +612,17 @@ func startBots(ctx context.Context, stack *krillStack) botStatus {
 				klog.Error("web (reef) error", "error", err)
 			}
 		}()
+		// Genuine feed presence: observe the social feed and engage selectively
+		// (like/comment/reply) when it's truly warranted. Shares ctx, llm, brain.
+		if stack.cfg.Feed.Enabled {
+			fw := feed.NewFeedWatcher(stack.llm, stack.brain, stack.cfg.Feed)
+			go func() {
+				if err := fw.Start(ctx); err != nil {
+					klog.Error("feed watcher error", "error", err)
+				}
+			}()
+			klog.Info("feed watcher enabled", "agent", reef.AgentID())
+		}
 		s.WebOK = true
 		s.web = wbBot
 		klog.Info("web (reef) bot started", "agent", reef.AgentID())
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -26,6 +26,20 @@ type Config struct {
 	Log      LogConfig      `yaml:"log"`
 	Doctor   DoctorConfig   `yaml:"doctor"`
 	TUI      TUIConfig      `yaml:"tui"`
+	Feed     FeedConfig     `yaml:"feed"`
+}
+
+// FeedConfig tunes how the agent genuinely engages with the Reef social feed:
+// it observes every post but acts on few. Threshold + budget are the levers that
+// keep engagement deliberate (selective) rather than spammy (forced). Per-agent
+// personalities set different values (Mini-Krill lively, Lab-Krill selective).
+type FeedConfig struct {
+	Enabled         bool    `yaml:"enabled"`           // master switch
+	Threshold       float64 `yaml:"threshold"`         // min appraisal interest (0-1) to engage
+	BudgetPerHour   int     `yaml:"budget_per_hour"`   // max engagements (likes+comments) per hour
+	PollIntervalSec int     `yaml:"poll_interval_sec"` // how often to scan for new posts/comments
+	MaxReplyDepth   int     `yaml:"max_reply_depth"`   // cap on agent<->agent reply chains (ping-pong guard)
+	Channels        string  `yaml:"channels"`          // optional comma-separated channel filter ("" = all)
 }
 
 // AgentConfig controls the main krill agent behaviour.
@@ -270,6 +284,16 @@ func DefaultConfig() *Config {
 		TUI: TUIConfig{
 			Theme: "ocean",
 		},
+		// Mini-Krill is the lively/playful participant: a low bar and a generous
+		// budget, so it comments and reacts more freely than Lab-Krill.
+		Feed: FeedConfig{
+			Enabled:         false, // opt-in; the owner turns it on per deployment
+			Threshold:       0.55,
+			BudgetPerHour:   12,
+			PollIntervalSec: 90,
+			MaxReplyDepth:   3,
+			Channels:        "",
+		},
 	}
 }
 
@@ -355,6 +379,18 @@ func fillDefaults(cfg *Config) {
 	}
 	if cfg.Brain.RecoveryTurns == 0 {
 		cfg.Brain.RecoveryTurns = 10
+	}
+	if cfg.Feed.Threshold == 0 {
+		cfg.Feed.Threshold = 0.55
+	}
+	if cfg.Feed.BudgetPerHour == 0 {
+		cfg.Feed.BudgetPerHour = 12
+	}
+	if cfg.Feed.PollIntervalSec == 0 {
+		cfg.Feed.PollIntervalSec = 90
+	}
+	if cfg.Feed.MaxReplyDepth == 0 {
+		cfg.Feed.MaxReplyDepth = 3
 	}
 	switch cfg.Brain.EmojiStyle {
 	case "none", "sparse", "playful":
--- a/internal/feed/watcher.go
+++ b/internal/feed/watcher.go
@@ -1,0 +1,390 @@
+// Package feed gives the agent a genuine presence in the Reef social feed.
+//
+// The design goal is GENUINE engagement, not forced: the agent observes every
+// post but acts on almost none. Whether to like, comment, or reply is an
+// APPRAISAL run through the local model, primed with the agent's own identity
+// and memory — selectivity is what reads as real. Three levers keep it from
+// becoming spam:
+//
+//   - a relevance THRESHOLD: only act when the model is genuinely interested;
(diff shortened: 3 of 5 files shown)


srvsngh99:
## Review - PR #48 (round 1)

Checked out `feature/feed-watcher` at head `195807c`. Ran:
- `go build ./...` - clean
- `go vet ./internal/feed/ ./internal/reef/` - clean
- `go test -race ./internal/feed/ ./internal/reef/ ./internal/config/` - all pass
- Read `internal/feed/watcher.go`, `internal/feed/watcher_test.go`, `internal/reef/client.go`, `internal/config/config.go`, `cmd/minikrill/commands.go` in context (memory key round-trip via `sanitizeKey`, `React` signature, `Brain`/`Memory` interfaces).

Solid overall: single watcher goroutine bound to ctx, clean exit on cancel, fail-safe appraisal parsing, correct sliding-window budget, self-engagement properly blocked. One substantive bug below.

### Substantive

1. `internal/feed/watcher.go:334` (`agentChainDepth`) - **infinite loop / goroutine hang on a cyclic parent chain (malformed data).** The walk has no visited-set and no ceiling on the number of hops. If two non-owner comments point at each other (`a.parent=b`, `b.parent=a`) - or any cycle of agent-authored comments - the `for cur.ParentCommentID != ""` loop never hits its exits (empty parent / parent-not-in-map / owner parent) and spins forever, pinning the watcher goroutine at 100% CPU; the FeedWatcher silently dies. The comment data is built from the hub's `GetComments` JSON, i.e. untrusted external input, so this is exactly the malformed-data case the guard is supposed to survive. Reproduced with a 2-comment cycle: the call hangs (2s test timeout fires; never returns).
   Fix: bound the walk. Either track visited IDs and break on revisit, or cap the hops at, say, `MaxReplyDepth+1` (we only care whether depth reaches the cap anyway):
   ```go
   func agentChainDepth(c reef.Comment, byID map[string]reef.Comment, me string) int {
       depth := 0
       cur := c
       seen := map[string]bool{cur.ID: true}
       for cur.ParentCommentID != "" {
           parent, ok := byID[cur.ParentCommentID]
           if !ok || isOwner(parent.Author) || seen[parent.ID] {
               break
           }
           seen[parent.ID] = true
           depth++
           cur = parent
       }
       return depth
   }
   ```
   Please also add a cycle case to `TestAgentChainDepth` so this can't regress.

### Nits

- `internal/feed/watcher.go:377` - `truncate` appends a literal unicode ellipsis `"…"` (U+2026) into runtime data, and slices `s[:n]` on a raw byte boundary. On multibyte UTF-8 content (emoji, accented text - common in agent posts) the byte slice can split a rune, producing a mojibake tail. It only affects memory notes and prompt text (the posted draft uses the full `a.Draft`), so impact is low, but prefer a rune-safe truncation and an ASCII `"..."` to keep the "hyphens/ASCII in code strings" house rule. Same U+2026 / em-dash usage in the prompt string literals at lines 125 and 219 - cosmetic, but they live in shipped strings.
- `internal/feed/watcher.go` doc-comment em-dashes (lines 6, 165, etc.) - comments only; flag for consistency with the no-dash convention.
- `internal/feed/watcher.go:227` (`engaged`) fails open on a Recall *error* (returns false, so the item gets re-appraised). Harmless in practice - re-engagement is still gated by threshold + per-hour budget, so it can't spam - but a transient FS error means a wasted appraisal call. Consider logging at debug when `err != nil`.
- `internal/feed/watcher.go:scan` always calls `GetFeed(ctx, 0, ...)` (since=0), so every cycle re-pulls the same 60-post window and relies entirely on memory dedup to avoid re-acting. Works, but you could pass a high-water `since` to cut the per-cycle scan and the memory Recall fan-out. Optional.
- `internal/reef/client.go` - `LikeComment` is exported but never called by the watcher. Fine to keep for the API surface; just noting it's currently dead from this PR's perspective.

### Verified good

- Build, vet, race tests all green on the targeted packages.
- `parseAppraisal` (watcher.go:357) is panic-safe: the `start>=0 && end>start` guard makes `s[start:end+1]` always in-bounds; a parse failure returns `ok=false` and the caller does NOT act (fail-safe = no engagement). The `no-json` test covers it.
- `budget` (watcher.go:299) - mutex usage correct; `prune` removes the monotonic-ordered stale prefix correctly; `allow` prunes-then-checks `< perHour`; cap is genuinely enforced (test confirms). Only one watcher goroutine writes it, so no real contention, but it's race-clean regardless.
- FeedWatcher goroutine is bound to ctx and exits cleanly: `Start` returns nil on `ctx.Err()` / `sleepCtx` cancel; per-call `appraise` uses a 60s child timeout.
- Self-engagement blocked on both paths: `scan` skips `p.Agent == w.me` for post actions; `handleReplies` skips `c.Author == w.me`.
- Dedup key scheme (`reef-eng:post:<id>` / `reef-eng:reply:<id>`) round-trips through `sanitizeKey` consistently for both Store and Recall, so it actually prevents re-acting on the same post/comment each cycle.
- New reef HTTP methods mirror the existing pattern: `X-Reef-Token` header, `ctx` on every request, `defer resp.Body.Close()`, drain + `status >= 300` error. No leaked bodies. `LikePost` correctly reuses `React(heart)`.
- `appraiseReply` raises the effective threshold by `0.12*depth` and the loop hard-stops at `>= MaxReplyDepth`, so chains decay and terminate (off-by-one is fine: depth is "agent levels above me", cap is inclusive).
- Per-cycle appraisal cap (`maxAppraisalsPerCycle`) correctly threaded through `scan` -> `handleReplies` via the `spent` accumulator.
- No secrets, sole-authored by Sourav Singh, no AI attribution in code.

### Verdict

One substantive bug (cyclic-chain hang) blocks. Fix `agentChainDepth` to bound the walk and add a cycle test; the rest is nits. CHANGES REQUESTED.

srvsngh99:
Addressed the round 1 blocking item in c9bc219: `agentChainDepth` now tracks visited comment ids, so a cyclic parent chain from malformed hub data stops the walk instead of hanging the watcher goroutine. Added `TestAgentChainDepthCycle` which fails by timeout if the walk ever spins. `go test -race ./internal/feed/` passes.

srvsngh99:
## Review - PR #48 (round 2)

Head SHA `c9bc219`. Ran on `feature/feed-watcher`:

- `go build ./...` - clean
- `go vet ./internal/feed/ ./internal/reef/` - clean
- `go test -race ./internal/feed/ ./internal/reef/ ./internal/config/` - all pass

### Round 1 follow-up - the cyclic parent chain

Verified fixed. `agentChainDepth` (internal/feed/watcher.go:334-353) now seeds a `visited` set with the starting comment id and bails the walk the moment it revisits a parent (watcher.go:339, 342-344, 349). A malformed a->b->a chain from the hub now terminates instead of spinning.

The regression test is real, not decorative. I pulled the visited-set guard out locally and `TestAgentChainDepthCycle` hung and tripped its 2s timeout ("agentChainDepth hung on a cyclic parent chain"); with the guard restored it passes. So the test actually exercises the hang path. Good.

### Substantive

None blocking.

### Nits

1. internal/feed/watcher.go:232 - `remember` swallows the `Store` error (`_ = ...`) and `engaged` fails-closed (a Recall error reads as "not engaged"). So if a memory write fails right after a successful comment/reply, the next scan re-appraises the same target and can post a second time. It is gated on memory infra actually failing, and the per-hour budget caps the blast radius, so it is not a blocker - but a one-line "warn on store failure" would at least surface the degraded mode in logs.

2. internal/reef/client.go - the GET helpers drain the body on the error path (`io.Copy(io.Discard, ...)` before returning the status error) but not on the happy path, relying on `json.Decoder` + `defer Close`. That is fine for keep-alive reuse here; just noting the asymmetry for consistency if you touch these again.

### Verified good

- No other unbounded loop or goroutine leak: `Start` checks `ctx.Err()` at the top of every iteration and sleeps via `sleepCtx` (cancellable); the watcher goroutine in commands.go exits cleanly on ctx cancel.
- Per-cycle appraisal cap (`maxAppraisalsPerCycle = 6`) vs budget: the budget is the hard rate lever (sliding hour window, mutex-guarded); the per-cycle cap only bounds model load. No spam path - every act goes through `bud.allow()` then `bud.record()`.
- `handleReplies` dedup is sound on the normal path: post-level via `post:<id>` and reply-level via `reply:<id>`, both written on every terminal outcome (act and ignore alike), so a re-scan skips already-handled items.
- Reef HTTP methods (GetFeed/GetComments/PostComment/LikeComment): all use `NewRequestWithContext`, `defer resp.Body.Close()`, check `StatusCode >= 300`, and drain the body. Correct.
- ctx cancellation mid-action: appraisal calls run under a 60s child context and post-cancel errors are suppressed (`if ctx.Err() == nil`), so shutdown does not spam warnings.

Solid. Shipping it.

VERDICT: APPROVED

srvsngh99:
Approved after 2 automated review rounds (round 1 caught a cyclic-chain hang in agentChainDepth, fixed in c9bc219; round 2 verified the fix and found nothing substantive). Merging.

PR #49, opened 2026-06-25
Proactive feed posting + per-account likes

srvsngh99:
## What the user asked for

> and agents are aware they post there anytime

(plus the matching reef change to real per-account likes)

## What changed

- `LikePost` now calls `POST /api/post/like` (real per-account, identity-bound like) instead of adding an anonymous heart reaction.
- Proactive posting: a slow second loop in the FeedWatcher lets Mini-Krill originate its own feed posts when it genuinely has something worth sharing. As the lively colony member it is chattier: every ~30m it considers posting, with a 0.5 interest bar, gated by the shared hourly attention budget. Most cycles it posts nothing.
- Config knobs added (`proactive_*`, default on, `post_channel=builds`).

Requires the reef social-actions PR deployed for the new like endpoint.

## How it was tested

`go build ./...`, `go test -race ./internal/feed/ ./internal/reef/ ./internal/config/` pass.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -40,6 +40,12 @@ type FeedConfig struct {
 	PollIntervalSec int     `yaml:"poll_interval_sec"` // how often to scan for new posts/comments
 	MaxReplyDepth   int     `yaml:"max_reply_depth"`   // cap on agent<->agent reply chains (ping-pong guard)
 	Channels        string  `yaml:"channels"`          // optional comma-separated channel filter ("" = all)
+	// Proactive posting: the agent originates its own feed posts when it
+	// genuinely has something worth sharing (gated by threshold + budget).
+	ProactiveEnabled     bool    `yaml:"proactive_enabled"`
+	ProactiveIntervalSec int     `yaml:"proactive_interval_sec"` // how often it considers posting
+	ProactiveThreshold   float64 `yaml:"proactive_threshold"`    // min interest to actually post
+	PostChannel          string  `yaml:"post_channel"`           // feed channel it posts to
 }
 
 // AgentConfig controls the main krill agent behaviour.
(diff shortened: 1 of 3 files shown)


srvsngh99:
## Review - PR #49 (round 1)

Head SHA: `43ba0e6` (branch `feature/social-actions`, base `main`)

Commands run:
- `go build ./...` - clean
- `go vet ./internal/feed/ ./internal/reef/` - clean
- `go test -race ./internal/feed/ ./internal/reef/ ./internal/config/` - all pass (race detector clean)

I read all three changed files in context plus the call site in `cmd/minikrill/commands.go` and the existing `PostIngest`/`LikeComment`/`React` patterns this PR mirrors.

### Substantive issues

None blocking. I went hunting for the obvious failure modes (goroutine leak, the race between the proactive loop and the reactive scan, a proactive firehose) and the implementation holds up:

- **Goroutine lifecycle** - `proactiveLoop` is launched exactly once in `Start` (watcher.go:70), gated by `ProactiveEnabled`, before the scan loop. `Start` itself is called once per process (commands.go:617-623), bound to the process `ctx`. The loop checks `ctx.Err()` each iteration and uses `sleepCtx`, which selects on `ctx.Done()` and returns false on cancel, so it exits cleanly with no leak. The first `sleepCtx(ctx, interval)` runs before the first `considerPost`, so it staggers and does not fire at boot. Correct on all three counts.
- **Concurrency** - the proactive loop and the reactive scan share only `w.bud`, which is mutex-guarded (`allow`/`record`/`prune`). `appraise` only reads `w.llm`/`w.brain`. `-race` is clean. No data race.
- **Spam safety** - gated by both the slow interval (default 1800s) AND `bud.allow()`. A no-say cycle does not burn budget (`record()` only fires after a successful `PostIngest`). Parse failure or model error -> `appraise` returns false -> early return, no post. Fail-safe.
- **LikePost** - mirrors `LikeComment` exactly: ctx-bound request, `Content-Type` + `X-Reef-Token` headers, `defer resp.Body.Close()`, drains the body, `status >= 300` check. No-ops on `!IsConfigured() || postID == ""`. Good.

### Nits (non-blocking)

1. config.go:299 / 405-413 - `ProactiveEnabled` defaults to `true` in `DefaultConfig()`, but `fillDefaults` has no entry for it (a bool can't be back-filled via the `== 0` idiom). So a user-supplied YAML that omits `proactive_enabled` gets Go's zero value `false`, i.e. proactive posting is OFF for existing on-disk configs even though the in-code default is ON. This is the conservative/fail-safe direction (and the whole feed is gated by `Feed.Enabled=false` anyway), so I am not blocking - just calling out the `DefaultConfig` vs `fillDefaults` divergence in case the intent was on-by-default for upgraders. A one-line comment noting "bool: opt-in via explicit config" would make the asymmetry intentional-looking.
2. watcher.go:269 `considerPost` - unlike `appraisePost`/`appraiseReply`, a proactive post writes nothing to memory, so there is no "I already said this" continuity record feeding future appraisals. That is by design here (the slow interval + the model's "do not repeat yourself" instruction carry the load), but it does mean repetition is purely model-discretion. Fine for now; worth a memory breadcrumb later if you see it echo itself.

### Verified good

- LikePost endpoint/body/headers/status-check match the established POST helpers.
- `PostIngest` is identity-bound via `AgentID()` in the payload; `PostChannel` ("builds") is passed straight through as the feed channel.
- `fillDefaults` covers the two non-bool new fields (`ProactiveIntervalSec`, `ProactiveThreshold`) and `PostChannel`.
- Branch is `feature/social-actions`, not main. No secrets, no AI attribution, no em/en dashes in added code or strings.

Verdict: ship it. The two nits are polish, not correctness.

srvsngh99:
Approved after 1 automated review round (race-clean, no substantive issues - goroutine lifecycle, budget concurrency, and spam-safety all verified). Merging.

PR #50, opened 2026-06-25
Respond when @tagged in the feed

srvsngh99:
When a post or comment @tags this agent (by its handle/id), the effective interest threshold drops to 0.2, overriding the normal selectivity and the depth-based ping-pong decay, and the model is told it was directly tagged. Untagged behaviour unchanged. Pairs with the reef @mention PR. Tests: go build + go test ./internal/feed/ pass.

### DIFF
--- a/internal/feed/watcher.go
+++ b/internal/feed/watcher.go
@@ -35,6 +35,11 @@ import (
 // pre-filter beyond this wait for the next cycle.
 const maxAppraisalsPerCycle = 6
 
+// mentionThreshold is the (low) interest bar applied when this agent is @tagged:
+// a direct address should almost always get a response, overriding the normal
+// selectivity and the depth-based ping-pong decay.
+const mentionThreshold = 0.2
+
 // FeedWatcher observes the feed and decides, genuinely, whether to engage.
 type FeedWatcher struct {
 	llm   core.LLMProvider
@@ -119,6 +124,16 @@ func (w *FeedWatcher) scan(ctx context.Context) error {
 // appraisePost asks the model whether this post genuinely warrants a like or a
 // comment, then acts if it clears the threshold and budget.
 func (w *FeedWatcher) appraisePost(ctx context.Context, p reef.FeedPost) {
+	// Being @tagged is a direct address: drop the bar so the agent almost
+	// always responds, and tell the model it was tagged.
+	effThreshold := w.cfg.Threshold
+	note := ""
+	if w.mentionsMe(p.Content) {
+		if effThreshold > mentionThreshold {
+			effThreshold = mentionThreshold
+		}
+		note = "\n\nNOTE: you were directly tagged (@" + w.me + ") in this post. You should almost certainly respond."
+	}
 	prompt := fmt.Sprintf(`A new post has appeared in the Reef feed (a private social feed for the Krill agent colony).
 
 POST by %s (@%s) in #%s:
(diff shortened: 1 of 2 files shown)


srvsngh99:
## Review - PR #50 (round 1)

Checked out `feature/mention-awareness` at `c22698b`. Ran:

- `go build ./...` - clean
- `go vet ./internal/feed/` - clean
- `go test -race ./internal/feed/` - ok (1.6s)

I read `internal/feed/watcher.go` and `watcher_test.go` in full around the changed regions, with attention to the loop backstops.

### Substantive issues

None. The boost is sound and the runaway-loop backstops are intact:

- The ping-pong hard stop (`depth >= cfg.MaxReplyDepth`, watcher.go:212) runs in `handleReplies` BEFORE `appraiseReply` is even called, and it does not consult mentions. So two agents @tagging each other forever still hit the hard depth cap and stop. The mention boost only lowers the *interest* bar inside `appraiseReply`; it cannot resurrect a chain past the cap.
- The shared hourly budget gate is unchanged: `w.bud.allow()` still guards `handleReplies` (watcher.go:216) and every `record()` site is unchanged. That remains the global ceiling on engagements per hour regardless of how many tags arrive.
- `effThreshold` is computed and actually used in both paths - `appraisePost` (set at :129-134, used at :155) and `appraiseReply` (set at :231-235, used at :256). No dead boost, no wrong gate. In `appraisePost` the `if effThreshold > mentionThreshold` clamp correctly avoids raising the bar if config is already below 0.2; in `appraiseReply` the unconditional `= mentionThreshold` correctly wipes the depth penalty, which is the stated intent.

### Nits (non-blocking)

1. watcher.go:436 - `mentionsMe` uses `strings.Contains(lower(s), "@"+lower(me))`, so a tag of a *different* agent whose handle is a superstring (`@minikrill2`, `@minikrillbot`, `@minikrill-helper`) would also boost this agent. Today the colony handles (deepkrill/labkrill/minikrill) do not collide and `me` is configurable via `REEF_AGENT_ID`, so this is latent, not a present bug; and the only effect is a lowered threshold + a "you were tagged" note - the depth cap and budget still bound everything, so it cannot cause harm. Acceptable over-match for now. If you want it airtight later, require a trailing word boundary (next char is end-of-string or not `[a-z0-9_-]`). Same applies to the left side (`foo@minikrill` matches), same low stakes.
2. watcher.go:229 - new code comment uses an em dash (`ping-pong guard) - UNLESS`). House style is hyphens in source; please swap. The two prompt-string em dashes flagged by a dash scan are pre-existing (unchanged by this PR), so leave those.
3. `TestMentionsMe` covers the `@mini` shorter-than-handle non-match (good), but not the `@minikrill2` over-match. Since the over-match is the accepted behaviour, no test is owed - just noting the test asserts the safe direction, not the loose one.

### Verified good

- No reintroduced agent<->agent infinite loop; depth cap + hourly budget both intact and mention-independent at the gate level.
- Boost correctly applied and consumed in both appraisal paths; untagged behaviour byte-for-byte unchanged.
- `@mini` (handle longer than tag) correctly does NOT match - the substring direction that matters is covered.
- No secrets, branch is `feature/mention-awareness` (not main), no AI attribution.

Verdict: approve. The over-match is a real but bounded edge worth a follow-up word-boundary tightening; it is not a blocker because the loop and budget backstops fully contain it.

VERDICT: APPROVED

srvsngh99:
Approved after 1 automated review round (loop safety intact - mention boost only lowers the interest bar, the depth cap and hourly budget still bound everything; no substantive issues). Merging.

PR #51, opened 2026-06-25
Use Krill engine as primary LLM (Ollama failover)

srvsngh99:
## What the user asked for

> why ollama we should be using krill and ollama is fallback to krill ... its Krill and not KrillLM anymore and no its not ollama is my own inference engine for local models ... let them take their own time (no reply timeout)

Decisions: auto-failover (Krill -> Ollama); Mini-Krill only (Lab-Krill stays on Claude); Krill owns the 12B, Ollama fallback uses a SMALL model.

## What changed

Mini-Krill was calling Ollama (`:11434`) even though its provider was *named* `krilllm` - it never actually used Krill. This wires it to **Krill** (`krillm serve` on `:57455`, the user's own OpenAI-compatible engine) as **primary**, with **Ollama** as an automatic **fallback**.

- `KrillProvider` - Krill's `/v1/chat/completions`, generous 10m timeout (cold MLX loads are slow; a background agent should take its time), refused-connection fails fast for quick failover.
- `FailoverProvider` - primary (Krill) then, on error, fallback (Ollama). Shutdown (ctx cancel) is never masked by a fallback. Stream fails over on outright error and errored first chunk.
- `provider.go` - krill/krilllm aliases build Krill primary + Ollama fallback; Krill URL/model env-overridable (`KRILLM_URL`/`KRILL_MODEL`). Fallback Ollama model is **small** (`gemma4:e2b`) on purpose: on a 16GB box only one 12B fits, so the fallback must coexist with Krill's 12B.
- Default provider is now `krill`; `OllamaConfig.FallbackModel` added.
- FeedWatcher appraisal drops its 60s deadline - only shutdown + the client backstop bound the call.

## How it was tested

`go build`, `go vet`, `go test ./...` green (new failover tests: primary-ok / fallback-on-error / respects-cancel / name; updated provider+config tests).

## NOT YET DEPLOYED

The `krillm` server is currently **wedged** (up 2+ days, chat completions hang) and needs a restart, and the 16GB box needs RAM headroom for the 12B. Deploy after Krill can serve a completion reliably. At deploy: set `~/.mini-krill/config.yaml` `llm.provider: krill`, reinstall the binary, reload the service.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -196,6 +196,10 @@ type OllamaConfig struct {
 	AutoInstall  bool   `yaml:"auto_install"`
 	AutoStart    bool   `yaml:"auto_start"`
 	DefaultModel string `yaml:"default_model"`
+	// FallbackModel is the model Ollama serves when it is the failover backend
+	// behind Krill. Kept SMALL on purpose so it can coexist in RAM with Krill's
+	// 12B on a 16GB box (a second 12B would not fit).
+	FallbackModel string `yaml:"fallback_model"`
 }
 
 // PluginsConfig for the skill registry.
@@ -257,8 +261,8 @@ func DefaultConfig() *Config {
 			AutonomyFloor: "act",
 		},
 		LLM: LLMConfig{
-			Provider:    "krilllm",
-			Model:       DefaultKrillLMModel,
+			Provider:    "krill", // Krill engine (:57455) primary, Ollama fallback
+			Model:       "gemma-4-12b",
 			Temperature: 0.7,
 			MaxTokens:   2048,
 		},
@@ -269,10 +273,11 @@ func DefaultConfig() *Config {
 			HeartbeatSec: 30,
 		},
 		Ollama: OllamaConfig{
-			Host:         "http://localhost:11434",
-			AutoInstall:  true,
-			AutoStart:    true,
-			DefaultModel: "gemma3:4b",
+			Host:          "http://localhost:11434",
+			AutoInstall:   true,
+			AutoStart:     true,
+			DefaultModel:  "gemma3:4b",
+			FallbackModel: "gemma4:e2b",
 		},
 		Plugins: PluginsConfig{
 			Dir: filepath.Join(dataDir, "skills"),
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -129,9 +129,11 @@ func (a AgentConfig) MarshalYAML() (interface{}, error) {
 	}, nil
 }
 
-// DefaultKrillLMModel is the primary model of the KrillLM provider —
-// Mini Krill's default, Ollama-backed local inference stack.
-const DefaultKrillLMModel = "gemma4:12b-mlx"
+// DefaultKrillLMModel is the primary model of the KrillLM provider. It must
+// match the model id the Krill engine (krillm serve on :57455) actually loads
+// and serves over its OpenAI-compatible API, otherwise requests name a model
+// the engine does not have and fail. The live engine serves "gemma-4-12b".
+const DefaultKrillLMModel = "gemma-4-12b"
 
 // IsLocalProvider reports whether the provider runs on the local Ollama
 // daemon (and therefore needs Ollama auto-start/health monitoring).
--- a/internal/config/config_test.go
+++ b/internal/config/config_test.go
@@ -10,8 +10,8 @@ import (
 func TestDefaultConfig(t *testing.T) {
 	cfg := DefaultConfig()
 
-	if cfg.LLM.Provider != "krilllm" {
-		t.Errorf("LLM.Provider = %q, want %q", cfg.LLM.Provider, "krilllm")
+	if cfg.LLM.Provider != "krill" {
+		t.Errorf("LLM.Provider = %q, want %q", cfg.LLM.Provider, "krill")
 	}
 	if cfg.Agent.PlanApproval != "auto" {
 		t.Errorf("Agent.PlanApproval = %q, want %q", cfg.Agent.PlanApproval, "auto")
@@ -25,8 +25,8 @@ func TestDefaultConfig(t *testing.T) {
 	if cfg.LLM.Model == "" {
 		t.Error("LLM.Model is empty, want a default model to be set")
 	}
-	if cfg.LLM.Model != DefaultKrillLMModel {
-		t.Errorf("LLM.Model = %q, want KrillLM primary model %q", cfg.LLM.Model, DefaultKrillLMModel)
+	if cfg.LLM.Model != "gemma-4-12b" {
+		t.Errorf("LLM.Model = %q, want Krill primary model %q", cfg.LLM.Model, "gemma-4-12b")
 	}
 	if cfg.LLM.Temperature != 0.7 {
 		t.Errorf("LLM.Temperature = %f, want 0.7", cfg.LLM.Temperature)
--- a/internal/feed/watcher.go
+++ b/internal/feed/watcher.go
@@ -324,13 +324,15 @@ type appraisal struct {
 // appraise runs one model call primed with the agent's identity and parses the
 // JSON verdict. A parse failure or model error is treated as "no engagement".
 func (w *FeedWatcher) appraise(ctx context.Context, userPrompt string) (appraisal, bool) {
-	callCtx, cancel := context.WithTimeout(ctx, 60*time.Second)
-	defer cancel()
+	// No artificial per-call deadline: a background agent should take as long as
+	// it genuinely needs to read a post and decide whether to reply. Only process
+	// shutdown (via ctx) and the model client's own connection backstop bound the
+	// call, so a slow or cold model load is not cut off mid-thought.
 	msgs := []core.Message{
 		{Role: "system", Content: w.brain.SystemPrompt()},
 		{Role: "user", Content: userPrompt},
 	}
-	resp, err := w.llm.Chat(callCtx, msgs, core.WithTemperature(0.3), core.WithMaxTokens(400))
+	resp, err := w.llm.Chat(ctx, msgs, core.WithTemperature(0.3), core.WithMaxTokens(400))
 	if err != nil || resp == nil {
 		if ctx.Err() == nil {
 			log.Warn("feed appraisal call failed", "error", err)
--- a/internal/feed/watcher.go
+++ b/internal/feed/watcher.go
@@ -35,6 +35,12 @@ import (
 // pre-filter beyond this wait for the next cycle.
 const maxAppraisalsPerCycle = 6
 
+// appraisalTimeout caps one appraisal call end to end. The failover stack drops
+// a hung primary to Ollama within its own bounded budget; this is the outer
+// guard so a single appraisal (primary plus any fallback) cannot stall the feed
+// loop for the model client's full 10m timeout.
+const appraisalTimeout = 3 * time.Minute
+
 // mentionThreshold is the (low) interest bar applied when this agent is @tagged:
 // a direct address should almost always get a response, overriding the normal
 // selectivity and the depth-based ping-pong decay.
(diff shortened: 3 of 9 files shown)


srvsngh99:
## Review - PR #51 (round 1)

Head SHA: `f86d609c38fec25c2cde453bcc2ebea34d70c788`

Commands run (default build, no `colony` tag):
- `go build ./...` -> PASS
- `go vet ./...` -> PASS
- `go test ./...` -> PASS (all packages ok)
- `go test -race ./internal/llm/... ./internal/config/... ./internal/feed/...` -> PASS

Read every changed/new file in full plus the surrounding call sites (`manager.go`, `provider.go`, `ollama.go`, `core/types.go`, `cmd/minikrill/main.go`). The core failover/cancel semantics and the new KrillProvider parsing are solid and well tested. The blocking problems are all downstream of one thing: the failover stack reports its Name() as `"krill+ollama"`, a string nothing else in the codebase recognizes.

### Substantive issues

1. **Runtime switch to krill bricks the next startup.** `manager.go:119` persists `m.cfg.LLM.Provider = newProvider.Name()`. For the new krill stack that is `"krill+ollama"` (`failover.go:80`). On the next launch `config.Load` -> `NewProvider("krill+ollama", ...)` hits the `default` case in `provider.go:81-83` and returns `unknown LLM provider "krill+ollama"`, which `main.go:138-140` turns into a hard `init LLM provider` failure. So once a user runs `/use krill` (or switches to it via the TUI), the app refuses to start until they hand-edit `config.yaml`. Fresh installs are fine (default `Provider: "krill"` loads), so this hides in testing.
   Fix: persist the canonical alias, not the composite display name - in `Switch`, write the normalized `provider` argument (`m.cfg.LLM.Provider = provider`) rather than `newProvider.Name()`. (Alternatively make `FailoverProvider.Name()` return the primary's canonical name, but that collides with the `"krill+ollama"` expectation baked into the new tests.)

2. **The provider catalog never marks the krill stack active.** `manager.go:173` gates the krill entry on `active.Provider == "krilllm"`. With the new default the active name is `"krill+ollama"`, so `ListProviders` shows *no* provider as active out of the box (this ships by default, no switch needed). Same stale-key assumption as #1.
   Fix: match the failover name, e.g. `IsActive: strings.HasPrefix(active.Provider, "krill")`, and reconcile the entry `Name`/aliases with what `Switch` persists.

3. **Krill model id drift between config and the catalog.** The engine is told to load `"gemma-4-12b"` (`config.go:264`, `krill.go:18` `KrillModelDefault`), but `manager.go:162` and the `/use` shorthands in `manager.go:341-345` still resolve to `DefaultKrillLMModel = "gemma4:12b-mlx"` (`config.go:134`). So the advertised model list is wrong, and `/use gemma12b` would send `"gemma4:12b-mlx"` to Krill's `/v1/chat/completions` instead of the id it actually serves. Pick one canonical Krill model id and use it in all three places.

4. **A wedged-but-connected Krill freezes the feed loop for up to 10 minutes with no failover.** `watcher.go appraise` dropped its 60s deadline and now passes the raw `ctx`, while `krill.go:74` sets a 10m client timeout. Failover only fires on an *error* (`failover.go:24`); a slow or hung-but-connected `krillm` returns no error, so the appraisal blocks the whole serialized feed loop for the full 10m per post and never drops to Ollama. This is exactly the failure state the PR description calls out ("chat completions hang"). The "refused fails fast" claim only holds for a TCP-refused dial; a connected-then-silent socket waits the entire timeout. Recommend bounding the primary call (a shorter primary timeout, or a watchdog that triggers failover) so a hung Krill degrades to Ollama instead of stalling the agent.

### Nits

- `krill.go:73` `Available` allocates a fresh `http.Client` on every probe; `OllamaProvider` keeps a reusable `healthClient`. Store one on the struct.
- `krill.go:106` `respBody, _ := io.ReadAll(...)` swallows the read error; a truncated body then surfaces as a misleading `decode krill response` error rather than the real I/O failure.
- `KrillProvider.Stream` is not actually streaming - it runs the full blocking `Chat` and emits a single terminal chunk, so `FailoverProvider.Stream` blocks reading the first chunk for the entire call. Fine for the interface contract, just no real token streaming through Krill primary.
- `krill.go` `if o.Temperature != 0` / `if o.MaxTokens != 0` means a caller cannot explicitly request temperature 0. Matches the existing Ollama convention, so pre-existing, not introduced here.

### Verified good

- Cancel semantics are correct and tested: `FailoverProvider.Chat`/`Stream` surface the primary error on `ctx.Err() != nil` and never burn a fallback call on shutdown.
- Genuine Krill errors (non-200, empty choices, malformed JSON, connection refused) all fail over to Ollama without panicking; bodies are drained and `defer resp.Body.Close()` is present in both `Chat` and `Available`.
- RAM coexistence holds by config: fallback is the small `gemma4:e2b`, so failover never loads two 12B models at once.
- `-race` clean on llm/config/feed; no goroutine leak beyond the standard "consumer must drain" stream pattern already used by Ollama.
- Hygiene: branch is `feature/krill-primary` (not main), no secrets, no AI attribution, no em/en dashes in added lines.

Solid direction and the failover unit tests are exactly the right ones. Items 1-4 need addressing before merge - 1 and 2 because the new default name breaks switch persistence and the active-provider UI, 3 for model-id correctness against the real engine, 4 because a hung krillm is the documented current state and it currently freezes the feed loop instead of failing over.

1. Switch persists `"krill+ollama"` -> next startup hard-fails on unknown provider (`manager.go:119`, `provider.go:81-83`, `main.go:138`)
2. Catalog `IsActive` never matches the new krill stack name -> no active provider shown by default (`manager.go:173`)
3. Krill model id drift: serves `gemma-4-12b` but catalog/aliases use `gemma4:12b-mlx` (`config.go:264` vs `config.go:134`, `manager.go:162,341-345`)
4. 10m timeout + dropped appraisal deadline -> a connected-but-hung Krill freezes the feed loop with no failover (`watcher.go appraise`, `krill.go:74`)

VERDICT: CHANGES_REQUESTED

srvsngh99:
Addressed items 1-4 in 0a594ed:

1. Startup brick: NewProvider now recognizes the failover stack's persisted name "krill+ollama" (provider.go), so the name round-trips and a restart after switching to/using the krill stack no longer hard-fails with "unknown LLM provider". Added a round-trip test (krilllm_test.go).
2. IsActive: the krilllm catalog entry is now gated through isKrillActive(), which matches the live default name "krill+ollama" (plus krill/krilllm), instead of the stale "krilllm"-only check (manager.go).
3. Model id drift: DefaultKrillLMModel is now "gemma-4-12b" to match what the engine actually serves (confirmed against /v1/catalog and /v1/status), aligning the catalog and aliases; kept the old "gemma4:12b-mlx" as a resolve alias (config.go, manager.go).
4. Hung-krill freeze: FailoverProvider runs the primary on a bounded (90s) derived context and falls over to Ollama on its deadline while still surfacing genuine shutdown via the parent ctx; the feed appraisal also regained a 3m outer deadline, so a wedged krillm can no longer stall the loop for the client's 10m timeout (failover.go, watcher.go).

go build / go vet / go test ./... all pass, including go test -race on internal/llm and internal/feed.

srvsngh99:
## Review - PR #51 (round 2)

Checked out `feature/krill-primary` at HEAD `0a594eddc4af3e5d9e306b5a5c4075253dc20dbd` (fix commit `0a594ed`), base `main`.

Commands run:
- `go build ./...` -> ok (rc=0)
- `go vet ./...` -> ok (rc=0)
- `go test ./...` -> all packages ok
- `go test -race ./internal/llm/... ./internal/feed/...` -> ok
- New tests verbose: `TestKrillProviderNameRoundTrips`, `TestFailoverDropsHungPrimary`, `TestFailoverRespectsCancel`, `TestNewProviderKrillLM`, `TestKrillModelOverride` -> all PASS
- `curl -s http://127.0.0.1:57455/v1/catalog` / `/v1/models` -> live engine serves alias `gemma-4-12b`

I did not write this code. Round 1 raised 4 substantive items; verifying each.

### Round 1 items - all genuinely fixed

1. STARTUP BRICK (failover name not recognized) - FIXED. `internal/llm/provider.go:44` adds `"krill+ollama"` (plus `krill`/`krilllm`/`krill-lm`/`krill_lm`) to the recognized case. `Switch` persists `newProvider.Name()` (`manager.go:119`), which for this stack is `FailoverProvider.Name()` = `"krill+ollama"` (`failover.go:107`), and that exact string now feeds back through `NewProvider` cleanly. `TestKrillProviderNameRoundTrips` (krilllm_test.go) proves the round trip. I checked the other persisted-name paths too: ollama/codex/claude/openai/anthropic/google all persist a `Name()` that `NewProvider` already recognizes, and the pre-PR `"krilllm"` value still boots. No remaining brick path.

2. IsActive gate - FIXED. `isKrillActive()` (`manager.go:138-144`) matches all three live names and is used at `manager.go:185`. With the running stack reporting `"krill+ollama"` the `krilllm` catalog row now lights up. It only matches the krill family, so it does not falsely flag ollama/codex/etc.

3. MODEL ID DRIFT - FIXED. `config.DefaultKrillLMModel = "gemma-4-12b"` (`config.go:136`); `DefaultConfig` sets `Model: "gemma-4-12b"` (`config.go:267`). The krill case passes `cfg.Model` straight into `NewKrillProvider` (`provider.go:54-58`), so the request body names `gemma-4-12b`, which the live `/v1/catalog` and `/v1/models` confirm the engine actually serves. The back-compat alias `gemma4:12b-mlx -> KrillLMDefaultModel` (`manager.go:357`) resolves to `gemma-4-12b`, so it does not reintroduce drift. `KrillLMDefaultModel` is now `= config.DefaultKrillLMModel`, so the catalog list and aliases also read `gemma-4-12b`.

4. HUNG-KRILL FREEZE - FIXED. `FailoverProvider.Chat` bounds the primary on a derived 90s context (`failover.go:18,42-46`) and falls over on either an outright error or the bounded-timeout, while `if ctx.Err() != nil { return err }` (`failover.go:48-50`) still surfaces genuine parent cancellation without a fallback attempt. `watcher.appraise` adds the 3m outer deadline (`watcher.go:38-39,335-340`). Ordering is correct: 90s primary < 3m appraisal < 10m client backstop, so a wedged krillm drops to Ollama in ~90s and the feed loop can never stall for 10m. No goroutine leak from the bounded attempt - the primary call is synchronous, so on pctx timeout `http.Client.Do` returns and the body is released by the transport; there is no double-call of the primary. `TestFailoverDropsHungPrimary` and `TestFailoverRespectsCancel` cover both paths.

No new regression found. The bounded-context shadowing in `appraise` is safe because cancellation still propagates from the real parent to the derived context.

### Nits
- `internal/llm/krilllm.go`: `KrillLMProvider` / `NewKrillLMProvider` are now dead code - nothing instantiates them after `provider.go` switched to `NewKrillProvider`. The file still earns its keep only via the `KrillLMDefaultModel` constant. Worth deleting the struct+constructor in a follow-up to avoid two "krill" provider types in the package.
- `FailoverProvider.Stream` (`failover.go:64-88`): on the first-chunk-error path it returns the fallback stream without draining the primary channel. Harmless for every current provider (krill/ollama emit a single terminating chunk, so the channel is already closed), but a future multi-chunk primary would leak its producer goroutine. Watcher uses `Chat`, not `Stream`, so no live impact.
- Fallback model default `gemma4:e2b` (`config.go:281`, `provider.go:66`) is an Ollama tag, separate from the krillm `gemma-4-e2b` alias - just confirm that tag is actually pulled on the target box so the failover path can answer.

### Verified good
- Round-trip safety across every persistable provider name.
- Model id sent to krillm matches the live engine catalog.
- Failover ordering (90s / 3m / 10m) and no-mask-on-shutdown semantics.
- `go build`, `go vet`, `go test`, `-race` all clean.
- Hygiene: on `feature/krill-primary` (not main); no em/en dashes in added lines; sole-authored by Sourav Singh, no AI attribution; no secrets (env reads only).

All four round-1 items are genuinely fixed with tests. Nits are non-blocking.

VERDICT: APPROVED

srvsngh99:
Approved after two automated review rounds. Round 2 confirmed all four round-1 fixes: the persisted provider name now round-trips through NewProvider (no startup brick), isKrillActive detects the new default, the catalog and aliases align with the gemma-4-12b the engine actually serves, and a hung krillm fails over within a bounded window without masking genuine cancellation. krillm single-owner is already live. Merging.

PR #52, opened 2026-06-26
Mini-Krill: semantic social memory + colony build-tag isolation

srvsngh99:
## What
Adds Mini-Krill's semantic social memory and isolates colony-only features from the public release via build tags.

- Reef feed + Chroma social memory are now colony-only, gated behind `//go:build colony`. The public release stays lean: FileMemory, no feed, no Chroma dependency.
- Colony builds (`make install-colony`, i.e. `go install -tags colony ./cmd/minikrill`) add `internal/socialmem` and wire the feed watcher's recall-before-comment / remember-after-engage via the shared bge embedder served by krillm.
- `maybeStartFeed` is a no-op in public builds (`commands_nocolony.go`) and launches the watcher in colony builds (`commands_colony.go`).

This was raised after PR #51, so the branch carries a merge of main; the merge keeps #51's bounded appraisal timeout and failover alongside this PR's social recall/remember.

## What changed (review fixes folded in)
- The one-time FileMemory to Chroma migration writes its "migrated" marker only when every entry succeeds; on failure it leaves FileMemory intact so a later boot retries (no data loss on a first-boot outage).
- Added a `make install-colony` target so the colony agent is built with the feed + social memory compiled in; the public `make install` intentionally excludes them.
- Social-memory writes are redacted (token/key/password/bearer/sk-/long-hex) before embed or persist, and the collection is retention-capped.

## How it was tested
- Both configs compile, vet, and test clean: `go build/vet/test ./...` (public) and `... -tags colony ./...` (colony); `go test -race -tags colony ./internal/socialmem ./internal/feed ./internal/brain` clean.
- Isolation verified: public binary has 0 feed/socialmem symbols; colony binary has them (socialmem 34, feed 26).
- Gated live socialmem test passes against the live Chroma + embedder.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,10 +1,13 @@
+# NOTE: The public build (`make install`) intentionally EXCLUDES the Reef feed
+# and the Chroma-backed social memory. A COLONY agent must be installed with
+# `make install-colony` (-tags colony) to compile those features in.
 VERSION := $(shell cat VERSION)
 BINARY  := minikrill
 LDFLAGS := -s -w -X github.com/srvsngh99/mini-krill/internal/core.Version=$(VERSION)
 
 .DEFAULT_GOAL := build
 
-.PHONY: build test lint clean install docker docker-down release fmt vet check help
+.PHONY: build test lint clean install install-colony docker docker-down release fmt vet check help
 
 ## build: Compile the binary to ./dist/minikrill
 build:
@@ -24,10 +27,15 @@ lint:
 clean:
 	rm -rf dist/
 
-## install: Install the binary via go install
+## install: Install the public binary (no feed / social memory) via go install
 install:
 	CGO_ENABLED=0 go install -ldflags="$(LDFLAGS)" ./cmd/minikrill
-	@echo "Installed $(BINARY) $(VERSION)"
+	@echo "Installed $(BINARY) $(VERSION) (public build)"
+
+## install-colony: Install the COLONY binary WITH feed + social memory (-tags colony)
+install-colony:
+	CGO_ENABLED=0 go install -tags colony -ldflags="$(LDFLAGS)" ./cmd/minikrill
+	@echo "Installed $(BINARY) $(VERSION) (colony build: feed + social memory)"
 
 ## docker: Build and start all containers
 docker:
--- a/README.md
+++ b/README.md
@@ -102,6 +102,11 @@ cd mini-krill
 go build -o minikrill ./cmd/minikrill
 ```
 
+> Note: `make install` (and the commands above) build the public binary, which
+> intentionally excludes the Reef feed and the Chroma-backed social memory. A
+> colony agent must be installed with `make install-colony` (`-tags colony`) to
+> compile those features in.
+
 See [docs/INSTALL.md](docs/INSTALL.md) for detailed setup including binary releases, Docker Compose, and PATH configuration.
 
 ---
--- a/cmd/minikrill/commands.go
+++ b/cmd/minikrill/commands.go
@@ -19,7 +19,6 @@ import (
 	"github.com/srvsngh99/mini-krill/internal/config"
 	"github.com/srvsngh99/mini-krill/internal/core"
 	"github.com/srvsngh99/mini-krill/internal/doctor"
-	"github.com/srvsngh99/mini-krill/internal/feed"
 	"github.com/srvsngh99/mini-krill/internal/llm"
 	klog "github.com/srvsngh99/mini-krill/internal/log"
 	"github.com/srvsngh99/mini-krill/internal/ollama"
@@ -612,17 +611,11 @@ func startBots(ctx context.Context, stack *krillStack) botStatus {
 				klog.Error("web (reef) error", "error", err)
 			}
 		}()
-		// Genuine feed presence: observe the social feed and engage selectively
-		// (like/comment/reply) when it's truly warranted. Shares ctx, llm, brain.
-		if stack.cfg.Feed.Enabled {
-			fw := feed.NewFeedWatcher(stack.llm, stack.brain, stack.cfg.Feed)
-			go func() {
-				if err := fw.Start(ctx); err != nil {
-					klog.Error("feed watcher error", "error", err)
-				}
-			}()
-			klog.Info("feed watcher enabled", "agent", reef.AgentID())
-		}
+		// Genuine feed presence + semantic social memory are COLONY-ONLY: they
+		// compile in only under `-tags colony`, so the public Mini-Krill release
+		// stays lean (no Reef feed, no Chroma dependency). maybeStartFeed is a
+		// no-op in public builds. See commands_colony.go / commands_nocolony.go.
+		maybeStartFeed(ctx, stack)
 		s.WebOK = true
 		s.web = wbBot
 		klog.Info("web (reef) bot started", "agent", reef.AgentID())
--- a/cmd/minikrill/commands_colony.go
+++ b/cmd/minikrill/commands_colony.go
@@ -19,7 +19,7 @@ func maybeStartFeed(ctx context.Context, stack *krillStack) {
 	// the other agents), embedded with the shared colony bge embedder. No-ops
 	// unless enabled and reachable.
 	// Unify storage: feed interactions and general memory share ONE collection
-	// per agent (the agent id) — its whole mind in one place — so feed recall can
+	// per agent (the agent id) - its whole mind in one place - so feed recall can
 	// surface general learnings and vice versa. Falls back to the agent id when
 	// no explicit collection is configured.
 	coll := stack.cfg.SocialMem.Collection
@@ -27,12 +27,13 @@ func maybeStartFeed(ctx context.Context, stack *krillStack) {
 		coll = reef.AgentID()
 	}
 	social := socialmem.New(reef.AgentID(), socialmem.Config{
-		Enabled:    stack.cfg.SocialMem.Enabled,
-		ChromaURL:  stack.cfg.SocialMem.ChromaURL,
-		EmbedURL:   stack.cfg.SocialMem.EmbedURL,
-		EmbedModel: stack.cfg.SocialMem.EmbedModel,
-		Collection: coll,
-		RecallK:    stack.cfg.SocialMem.RecallK,
+		Enabled:      stack.cfg.SocialMem.Enabled,
+		ChromaURL:    stack.cfg.SocialMem.ChromaURL,
+		EmbedURL:     stack.cfg.SocialMem.EmbedURL,
+		EmbedModel:   stack.cfg.SocialMem.EmbedModel,
+		Collection:   coll,
+		RecallK:      stack.cfg.SocialMem.RecallK,
+		RetentionCap: stack.cfg.SocialMem.RetentionCap,
 	})
 	if social.Enabled() {
 		klog.Info("social memory enabled", "collection", coll)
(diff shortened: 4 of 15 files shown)


srvsngh99:
## Review - PR #52 (round 1)

Head SHA: `faeb040d19dee510d025b7ef97d19243ffd7cf18` (merge of origin/main into feat/social-memory-colony-isolation). Branch checked out, `git pull --ff-only` clean, on branch (not main).

Commands run:

- `go build ./...` -> clean
- `go vet ./...` -> clean
- `go test ./...` -> all green (feed/socialmem absent from public, as intended)
- `go build -tags colony ./...` -> clean
- `go vet -tags colony ./...` -> clean
- `go test -tags colony ./...` -> all green (feed + socialmem compile and run; live tests skip without SOCIALMEM_LIVE=1)
- `go test -race -tags colony ./internal/socialmem/... ./internal/feed/... ./internal/brain/...` -> clean, no races

Isolation (built both binaries from ./cmd/minikrill, symbol/string counts):

| symbol | public nm / strings | colony nm / strings |
|---|---|---|
| socialmem | 0 / 0 | 27 / 46 |
| internal/feed | 0 / 0 | 26 / 32 |
| ChromaMemory | 0 / 0 | 8 / 9 |
| FeedWatcher | 0 / 0 | 17 / 20 |

The public binary carries ZERO feed/socialmem/Chroma symbols; the colony binary carries them. Every colony-only file is `//go:build colony` (socialmem/*.go, memory_chroma*.go, watcher.go, watcher_test.go, commands_colony.go) and each has a `!colony` complement where one is needed: memory_default.go (`!colony`) and commands_nocolony.go (`!colony`) provide the SAME signatures (`newDefaultMemory`, `maybeStartFeed`) so both builds compile. The Chroma/socialmem dependency is genuinely dropped from the public build, not just no-oped at runtime. Isolation feature works as designed.

Merge correctness (the round's crux): both files are semantically correct.
- config.go: net diff over main shows ONLY the new `SocialMemConfig` + default block; #51's LLM/provider config is untouched in main and preserved.
- watcher.go: #51's bounded `appraisalTimeout = 3 * time.Minute` const survived and `appraise()` still wraps each call in `context.WithTimeout(ctx, appraisalTimeout)` with the failover note intact; #52 added `social` recall/remember on top. No regression.

Below are the substantive issues, then nits.

### Substantive

1. **memory_chroma.go:176-196 - migration marker is written even when the migration fails, so a transient Chroma/embedder outage at first colony boot permanently orphans the legacy FileMemory.** `migrateFileMemoryOnce` loops `cm.Store(...)` (which hits Chroma over the network); if Chroma or the embedder is down, every `Store` returns an error, `migrated` stays 0, and the function STILL runs `os.WriteFile(marker, ...)` and logs "migrated 0 of N". On the next boot the marker exists, migration is skipped forever, and the agent's pre-existing memory is never carried into Chroma. The FileMemory JSON is left on disk but the colony build never reads it again. Fix: only write the marker when the migration actually completed - e.g. `if migrated == len(entries) { os.WriteFile(marker...) }` (and still write it when `len(entries)==0` so a fresh install isn't retried every boot). Returning/aborting without the marker on partial failure lets the next boot retry.

2. **Makefile:28 - `make install` does not pass `-tags colony`, so deploying the live Mini-Krill agent with it silently compiles OUT the feed watcher AND the social memory this PR adds.** The default `install` target runs `go install ... ./cmd/minikrill` with no build tag, producing the public (lean) binary. The live colony agent needs the colony build, so shipping it via `make install` would defeat the entire feature. Add a `make install-colony` target (`go install -tags colony ...`) and/or document in the PR/README that the colony agent must be built with `-tags colony`. Without this the feature is reachable in tests but not in the standard deploy path.

### Nits

- socialmem.go:97,129 use stdlib `log.Printf` ("[socialmem] ...") instead of the project's structured `internal/log`. With `SocialMem.Enabled` defaulting to true (config.go:331) and Chroma down, every feed engagement logs a remember failure - log spam on an unstructured logger. Prefer klog at debug level, or rate-limit.
- memory_chroma.go: the colony default memory unifies general memories and feed social rows in one collection (collection == agent id). `List`/`Search`/`Count` therefore also surface feed-interaction rows written by `rememberSocial`, which carry no `key` metadata, so `hitsToEntries` yields entries with an empty `Key`. Intended per the docstring, but worth confirming downstream consumers tolerate empty-key entries.
- Search/RankedSearch (memory_chroma.go:76-89) swallow Chroma errors and return `(nil, nil)` while Store/List/Count propagate. A Chroma outage looks like "no relevant memories" with no error surfaced. Acceptable for best-effort, but the error handling is inconsistent across the interface.
- No retention/eviction in ChromaMemory; `max` only bounds reads (List/All), writes are unbounded. Not a regression (FileMemory's `maxMem` is also never enforced for eviction), but the "~1k entries" assumption in the docstring is load-bearing and undocumented as a limit.
- No PII/secret redaction before storing feed/DM text verbatim into Chroma. Low risk for a private local colony, but the docstring advertises DM capture, so note it.
- Added comments and two runtime strings (memory RecallBlock header at socialmem.go:161 and the " - " separator at :169) use em dashes; house style is hyphens. Pre-existing style in this package, but new lines repeat it.

### Verified good

- Build-tag isolation is correct and complete; public build proven free of feed/socialmem/Chroma symbols, colony build has them; complementary `!colony` files provide matching signatures.
- nil-safe / best-effort throughout: `Store.Enabled()` guards on `s != nil`; Remember/Recall/RecallBlock no-op on a nil or disabled store; `rememberSocial` double-checks `w.social == nil`. A down embedder/Chroma never crashes the feed.
- Per-agent collection isolation: default collection `<agent>` / `<agent>_social`, agent stamped into metadata.
- Deterministic ids for dedup (`like:`/`comment:`/`reply:` + post/comment id) so Upsert is idempotent.
- All HTTP calls bound context with `WithTimeout` + `defer cancel()` and `defer resp.Body.Close()`; no goroutine/body leaks found; `-race` clean.
- `agentChainDepth` guards against cyclic parent chains with a visited set (no spin).
- core.Memory satisfied by both backends (`var _ core.Memory = (*ChromaMemory)(nil)`); colony/public swap is clean via `newDefaultMemory`.
- Merge preserved #51 (appraisalTimeout/failover, LLM config) and #52 (social + SocialMem config) in both touched files.
- Branch (not main), no new committed secrets, no AI attribution in code, `-race` clean.

Note (does not affect the code verdict): the PR BODY contains a "Generated with Claude Code" line - strip it before merge.

---

Two substantive items, both fixable in place. Requesting changes.

1. memory_chroma.go:176-196 - migration marker written on failure permanently orphans legacy FileMemory after a transient outage at first colony boot.
2. Makefile:28 - `make install` lacks `-tags colony`, so the standard deploy compiles out the feed + social memory; add `make install-colony` and/or document.

VERDICT: CHANGES_REQUESTED

srvsngh99:
Addressed the review items in 89f383c:

- Migration data-loss: migrateFileMemoryOnce now tracks an allOK flag and only writes the .chroma_migrated marker when every entry stored; on any failure it logs and leaves FileMemory intact so a later boot retries.
- Makefile: added install-colony (-tags colony) and a note that the public make install intentionally excludes the feed + social memory; mirrored the note in the README install section.
- Redaction + retention: secrets (token/key/password/bearer/sk-/long hex) are redacted at the Upsert chokepoint before anything is embedded or persisted, and the collection is trimmed to a config-driven retention cap (default 2000).
- Replaced em/en dashes with hyphens in the added colony lines.

Both build configs pass: go build/vet/test ./... and the same with -tags colony, plus go test -race -tags colony on socialmem/feed/brain.

srvsngh99:
## Review - PR #52 (round 2)

Checked out `feat/social-memory-colony-isolation` at HEAD `89f383c94ce7a0a7c48a82af86c19ad32462b488`.

Commands run and results:
- `go build ./...` / `go vet ./...` / `go test ./...` (public): clean, all packages pass.
- `go build -tags colony ./...` / `go vet -tags colony ./...` / `go test -tags colony ./...`: clean, all packages pass (socialmem + feed compile in).
- `go test -race -tags colony ./internal/socialmem ./internal/feed ./internal/brain`: pass, no races.
- Isolation (built both binaries from `./cmd/minikrill`, `go tool nm`):
  - public: socialmem symbols = 0, internal/feed symbols = 0
  - colony: socialmem symbols = 34, internal/feed symbols = 26
- gofmt clean on every file this PR touches.

Round-1 items re-verified:

1. DATA-LOSS migration (memory_chroma.go) - FIXED. `migrateFileMemoryOnce` (internal/brain/memory_chroma.go:175-209) now tracks `allOK`; on any `cm.Store` failure it sets `allOK = false`, logs, and `continue`s. The `.chroma_migrated` marker is written only inside the `if !allOK { return }` guard's fall-through (line 207), i.e. exclusively when every entry stored. FileMemory is opened read-only via `NewFileMemory` + `List`; nothing deletes or truncates it, so a later boot with Chroma up retries cleanly. No path marks done on partial failure. Empty-store case (no entries) correctly marks done (nothing to lose).

2. MAKEFILE (Makefile) - FIXED. New `install-colony` target (Makefile:36-39) runs `go install -tags colony -ldflags="$(LDFLAGS)" ./cmd/minikrill` - same ldflags, same target as public. Added to `.PHONY`. Public `install` (Makefile:31-34) is unchanged behaviourally (only the echo string + comment changed, no `-tags`). README + Makefile header note the split.

Parity items re-verified:
- Redaction: `redact()` (socialmem.go) is applied in `Upsert` (store_ops.go:27) BEFORE `embed` and before the `/upsert` POST. Both write paths funnel through Upsert: feed `Remember` -> Upsert, and `ChromaMemory.Store` -> `m.store.Upsert`. No bypass - Upsert is the only write. Regex is RE2 (no catastrophic backtracking) and covers key=value secrets, `sk-` keys, and 32+ hex blobs as claimed.
- Retention: `trim()` sorts ascending by `ts` and deletes `recs[:len(recs)-cap]` - oldest dropped, newest `cap` kept, off-by-one correct. Triggered every 64 writes via `atomic.AddInt64(&s.remCount,1)%64==0`, bounding growth to ~cap+64. `RetentionCap` is config-driven (default 2000). trim reads a snapshot via `All` then deletes by id (idempotent), so concurrent writes are not corrupted; race test passes.

### Nits
- Retention ordering in the unified brain collection: `ChromaMemory.Store` (memory_chroma.go:55-65) writes `created_at`/`accessed_at` but no `ts` field, while `trim` orders by `ts`. In the agent-id collection (general memories + feed memories share it), general entries evaluate `ts=0` and sort as "oldest", so if the collection ever exceeds `RetentionCap` (2000) the agent's core memories would be evicted before feed chatter. Only bites past 2000 rows and is colony-only; consider falling back to `created_at` when `ts` is absent.
- Two trims can overlap if writes cross 64-boundaries in quick succession; harmless (Delete-by-id is idempotent) but a `CompareAndSwap` guard would avoid the redundant scans.

### Verified good
- Build-tag isolation is real and total: public binary carries zero feed/socialmem symbols; colony carries them. `maybeStartFeed` no-op shim (commands_nocolony.go) vs colony impl is clean.
- Best-effort/nil-safe design preserved: disabled or unreachable Chroma/embedder silently no-ops.
- No em/en dashes in added lines; no AI attribution in added lines; no new committed secrets; branch is not main.

VERDICT: APPROVED

srvsngh99:
Approved after two automated review rounds. Round 2 confirmed both round-1 fixes plus parity items: the FileMemory to Chroma migration marks itself done only when every write succeeds (FileMemory left intact on failure, no data loss), make install-colony builds the agent with the feature compiled in, social writes are redacted before embed/persist, and the collection is retention-capped. Build-tag isolation verified (public binary has 0 socialmem/feed symbols, colony has them), and the merge keeps #51's bounded timeout and failover. Merging.

